Issue #87, opened 2016-11-04
Glucose data is removed from Healthkit when it doesn't have to


beached:
The medtronic cgm users Healthkit glucose data is purged if older than 3hrs on every store of a glucose value.  On Dex this was wanted to prevent duplicate data.  

I have commented it out and I am testing on a medtronic only system right now, but I would like to see at the very least the data left there for Medtronic CGM's.

CrushingT1D:
I see no reason to delete old BG values that Loop has written to healthkit. Current versions of iOS allow for settings in HealthKit where users can determine the priority they'd like to have their data used and does deduping based upon that.

This issue will also make it so that users are not required to turn on Health integration inside their Dexcom app.

beached:
I have a PR ready to go

apabari:
I haven't been able to figure out how to add this functionality to my loop app.  Wondering, will this PR / functionality be added to the main branch of loop?

CrushingT1D:
Try looking in CarbKit in LoopKit

beached:
My fork of loopkit has the one line commented out that prevents removal.  It works now but @ps2 had a better way if i remember

CrushingT1D:
Any update on this one @beached @ps2 ?

ps2:
In loop, you can set `glucoseStore.managedDataInterval = nil` and your data will not be purged.

iMaces:
Still having issues making this work.
Is it just copy/paste the code above or is there some extra work needed?

PR #88, opened 2016-11-04
commented out called to purgeOldGlucoseSamples. 

beached:
This is a fix for https://github.com/LoopKit/LoopKit/issues/87

### DIFF
--- a/GlucoseKit/GlucoseStore.swift
+++ b/GlucoseKit/GlucoseStore.swift
@@ -124,7 +124,7 @@ public final class GlucoseStore: HealthKitSampleStore {
                     let sortedGlucose = glucose.sorted { $0.startDate < $1.startDate }
 
                     self.sampleDataCache.append(contentsOf: sortedGlucose)
-                    self.purgeOldGlucoseSamples()
+                    //self.purgeOldGlucoseSamples()
 
                     if let latestGlucose = sortedGlucose.last, self.latestGlucose == nil || self.latestGlucose!.startDate < latestGlucose.startDate {
                         self.latestGlucose = latestGlucose


beached:
I have been running this for 2 days with a Medtronic enlite setup and no issues.

GlucoseKit/GlucoseStore.swift
```diff
@@ -124,7 +124,7 @@ public final class GlucoseStore: HealthKitSampleStore {
                     let sortedGlucose = glucose.sorted { $0.startDate < $1.startDate }
 
                     self.sampleDataCache.append(contentsOf: sortedGlucose)
-                    self.purgeOldGlucoseSamples()
+                    //self.purgeOldGlucoseSamples()
```
ps2: For G4/G5 cases (non enlight), we will still want to purge old data. The comment about source preference doesn't necessarily apply to glucose. It's intended for exercise data.  Any other app that queries glucose data will get duplicates.  Dexcom is shipping backfill-enabled transmitters soon, so it's best to let its own app put data in HealthKit. 

There are (at least) two possible routes for dealing with this:

1) Set managedDataInterval public ivar to null in the case of enlite, which will prevent purging.  If we do this, the recentSamplesPredicate method should be changed make startTime non-optional (the only call to it already supplies startTime).

2) Add another metadataKey to denote long/short term storage, and modify the addGlucose* methods to accept a flag.  The different places we add data (enlite, g4, g5) would set the flag.  And the purge routine would only purge entries with the flag set.

beached: The device field in addGlucose wouldn't be of use would it?  Then if the device is a known to backfill data purge, otherwise don't purge?

ps2: Going to close this.  A different approach, that allows dexcom users to continue having dexcom manage its HK data, would be acceptable.

ps2:
Going to close this. A different approach, that allows dexcom users to continue having dexcom manage its HK data, would be acceptable.

PR #91, opened 2017-01-02
Turn on whole module optimization, as recommended by Xcode

bharat:


bharat:
apparently this can cause crashes in Loop due to symbols being incorrectly optimized away. Hold off on pulling this until I can do an end-to-end test.

bharat:
ok - I've run through the basics on the app and found/fixed the one issue where the optimization was getting rid of a reachable path. I'll file a bug against Apple for that issue but in the meantime I think this is ready to merge. We should get it into the dev branch of Loop ASAP to get more eyes on it.

PR #93, opened 2017-01-02
Possible fix for issues with reading carbs from healthkit

ps2:
Remove some unnecessary async behavior that may be contributing to a race condition.


ps2:
Actually fixed with #101

PR #96, opened 2017-02-08
Add validation for Carb Entry Absorption time (less than 999)

sharpfive:
Don't allow absorption time entries over 999. Basic validation from the Carb Entry Edit VC.

https://github.com/LoopKit/Loop/issues/312

@ps2


ps2:
I can see it popping up a window to suggest < 999 minutes for CA when I enter large values, but when I do enter a valid value, it doesn't seem to be saving.

sharpfive:
Fixed. There is code checking the sender of the Unwind Segue. If it's not the SaveButton it's assumed to be a Cancel.

PR #105, opened 2017-03-09
LoopKit updates

ps2:
- Improving API clarity: result types, requiring date values, adjusting naming
- Removing caches that are better-applied upstream
- GlucoseStore: better management of HealthKit errors
- DoseStore: supporting longer dose & IOB queries for Loop's status graphs
- Changing the bundle identifiers to `com.loopkit.`


ps2:
wrong branch

PR #121, opened 2017-04-24
Feature/disregard primes using event history

sharpfive:
Don't use reservoir values if a Prime Pump Event has been detected during the Insulin On Board time.

### DIFF
--- a/InsulinKit/PumpEvent.swift
+++ b/InsulinKit/PumpEvent.swift
@@ -119,17 +119,17 @@ extension PumpEvent {
 
             return DoseEntry(type: type, startDate: startDate, endDate: endDate, value: value, unit: unit, managedObjectID: objectID)
         }
-        set {
-            guard let entry = newValue else {
-                return
-            }
-
-            type = entry.type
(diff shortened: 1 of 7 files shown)


InsulinKit/PumpEvent.swift
```diff
@@ -119,16 +119,22 @@ extension PumpEvent {
...
-            type = entry.type
-            startDate = entry.startDate as Date
-            endDate = entry.endDate as Date
-            value = entry.value
-            unit = entry.unit
+    }
+    
+    func setValuesWith(pumpEventType: PumpEventType?, dose: DoseEntry?) {
+        // This could also be done inline in DoseStore.addPumpEvents
+        // I'm OK with either place since it's only used once
```
ps2: I think this is harder to read than the previous version (inline is less work to comprehend).  And it's more LOC.

sharpfive: Fine with me. I kept this here because it was an adaption of the previous setter in the Event extension.

ps2:
Continuing work in #122

PR #132, opened 2017-05-20
Make CarbStore's defaultAbsorptionTimes mutable

mpangburn:


ps2:
maximumAbsorptionTimeInterval should be updated when defaultAbsorptionTimes is, and that in turn means that the observerQueries and the queryAnchors need to be reset.

CarbKit/CarbStore.swift
```diff
@@ -110,7 +110,12 @@ public final class CarbStore: HealthKitSampleStore {
     public var carbRatioSchedule: CarbRatioSchedule?
 
     /// A trio of default carbohydrate absorption times. Defaults to 2, 3, and 4 hours.
-    public let defaultAbsorptionTimes: DefaultAbsorptionTimes
+    public var defaultAbsorptionTimes: DefaultAbsorptionTimes {
+        didSet {
+            self.maximumAbsorptionTimeInterval = defaultAbsorptionTimes.slow * 2
+            createQueries()
```
ps2: Any existing queries should be stopped with healthStore.stop(query).  And self.queryAnchors should be cleared.

ps2:
These changes are not compiling. You need to loop through the queries, and maximumAbsorptionTimeInterval is a let variable.  We should probably make maximumAbsorptionTimeInterval a computed property, to avoid duplicating the code to set it.

mpangburn:
Interesting--I assumed Xcode would catch maximumAbsorptionTimeInterval being immutable and stop me from trying to reassign its value. Hopefully this one does the trick..

ps2:
Please at least build your changes before submitting...

mpangburn:
I do! I'm not having any issues building on my laptop: https://i.imgur.com/ieRvGq6.png (I've also tried cleaning the project and restarting Xcode, still builds without a problem.)
I apologize for making such a poor first impression, but I have no idea why the Travis build is failing.. and the detail isn't providing much insight from my perspective. Any idea what's causing the failure?

ps2:
No worries. Thanks for getting it passing!  If, along this journey of starting to code in Loop, you find things that might help others get started with contributing, that would be a great addition to the loop docs.

mpangburn:
I'll be sure to keep track of those things I encounter as I'm learning. Thanks for all your help!

ps2:
This is looking good, @mpangburn!  You were doing well to follow precendent and put the UI elements in LoopKit, but that's a leftover pattern from when Nate was thinking people would be building their own custom apps on top of LoopKit, instead of modifying a fully built app like Loop.  It's easier to work on the UI elements if they're in Loop, so it would probably be best to move them over there.

CarbKit/UI/DefaultAbsorptionTimesTableViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        case fast
+        case medium
+        case slow
+        
+        static let count = 3
+        
+        var labelText: String {
+            switch self {
+            case .fast:
+                return "🍭 Fast"
```
ps2: localization

CarbKit/UI/DefaultAbsorptionTimeTableViewCell.swift
```diff
@@ -0,0 +1,66 @@
...
+        if selected && timePicker.isEnabled {
+            let closed = timePicker.isHidden
+            
+            timePicker.isHidden = !closed
+            timePickerHeightConstraint.constant = closed ? timePickerExpandedHeight : 0
+        }
+    }
+    
+    @IBAction func timeChanged(_ sender: UIDatePicker) {
+        timeLabel.text = "\(Int(time.minutes)) min"
```
ps2: Should be using localization support here; like DateComponentsFormatter.

mpangburn:
For the time being I've left the UI elements in LoopKit (mostly since the UITableViewController is being created programmatically, it seems funny to drop one in a storyboard just to include the cell), but if I end up doing a lot more on them I'll move them over to Loop for convenience.

ps2:
Updated a few parts of this, and pushed it to dev as #145.  Thanks @mpangburn!

PR #140, opened 2017-06-23
Dynamic carb absorption

ps2:
Code to support implementation of dynamic carbs in Loop.  This code has a backwards compatible path that should still support the way carb absorption is used currently in Loop/dev.  There are related Loop changes that will be added as a separate Loop branch in the near future.


jmatheson:
#like

jessiepusateri:
👍  Very exciting!

Issue #38, opened 2016-07-09
Use structs for DailyValueSchedule types


loudnate:
When I was originally ad-hoc designing these types, I saw a lot of overlap but I at the time I could only articulate it using inheritance. Considering the properties are all immutable and there's no reliance on sharing references to a single instance currently in Loop, the inheritance pattern is the only blocker.

ps2:
Fixed in #155

PR #155, opened 2017-07-29
iOS 10 APIs, single HKHealthStore, daily schedules as value types

ps2:
* Removing the Core Data "Fetchable" protocol extensions in favor of iOS 10 APIs
* Support sharing a single HKHealthStore across all uses
* Changing the daily schedules to be value types for better change semantics
* Refactoring the GlucoseRangeSchedule to support multiple override types
* Improved initial UX in GlucoseRangeScheduleTableViewController


ps2:
Resolves #38

PR #189, opened 2017-12-26
add initialDelay to exponential insulin model

kenstack:


kenstack:
add initial delay term to exponential insulin model.  If initialDelay is set to zero it defaults to current model.

PR #199, opened 2018-02-18
LoopKit support for remote temp targets

Kdisimone:
Allows for .remoteTempTarget for Nightscout careportal temp targets in Loop


Kdisimone:
Thank you to @kenstack for the work on this!  We've been testing it for some time and have enjoyed the feature.

PR #246, opened 2018-11-04
The case for fat and protein.

rsilvers129:
rsilvers FPU modifications.

Adds in support for fat and protein.

See Wiki: 

https://github.com/rsilvers129/Loop/wiki/Fat-Protein-Unit-Branch-(FPU-dev)-of-Loop!-With-IRC-and-Spike.

See video:

https://www.youtube.com/watch?v=OvpRUlD0acU

### DIFF
--- a/Common/LocalizedString.swift
+++ b/Common/LocalizedString.swift
@@ -19,3 +19,4 @@ func LocalizedString(_ key: String, tableName: String? = nil, value: String? = n
         return NSLocalizedString(key, tableName: tableName, bundle: FrameworkBundle.main, comment: comment)
     }
 }
+// This is a useless comment as a carthage test.
--- a/LoopKit.xcodeproj/project.pbxproj
+++ b/LoopKit.xcodeproj/project.pbxproj
@@ -1435,7 +1435,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0720;
-				LastUpgradeCheck = 1000;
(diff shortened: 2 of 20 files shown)


birdbysnow:
Hi @rsilvers129,  apologies if this is not the right place to reach out or if the following question has a really obvious answer. I’m new to Loop and github and literally worked through the helloworld tutorial just yesterday. I’m super excited, I’m also still a little lost.

I’ve been reading the FPU branch discussions on different platforms and I’d really like to try it out. Only, I have an Omnipod and I saw that LoopDocs recommends using the omnipod-testing branch for stability.

Any chance you can tell me or point me to some documentation on how go about adding the FPU modifications to a download from omnipod-testing rather than master or dev?

mitchellhenke:
@birdbysnow I’m aiming to have a version out this weekend. Discussion and any release notifications are here: https://loop.zulipchat.com/#narrow/stream/144182-development/topic/FPU

ps2:
The Loop model should be updated to track fat and protein separately, and store those nutritional values separately to HealthKit.  The underlying math could still use "equivalency" units, if desired, but those equivalencies should be tracked separately until it's time for the algorithm to produce a single glucose effect vector.

Usage of fat and protein from HealthKit should be configurable via the LoopKit api, so we can control it via a feature flag in Loop.

I'm going to close this for now, but will provide review on a new submission aimed at tackling this as described above.

PR #250, opened 2018-12-18
Mock pump and CGM managers

mpangburn:
Features:
- Pump simulator with controllable reservoir and battery, ability to force errors on bolus/temp/suspend/resume.
- CGM simulator with controllable glucose production model (constant or sine curve) and composable effects (glucose noise, random outliers, random errors).
- Simulator configuration via settings; ability to clear out simulator data from HealthKit.

Notes:
- Strings are deliberately not localized; this is a development tool and not for production.
- Bolus and suspend states don't make use of transitional "bolusing" or "suspending" states.


mpangburn:
Reopening on tidepool-org/LoopKit

PR #254, opened 2019-01-11
Avoid double counting boluses when using reservoir

ps2:


ps2:
This adds a possible undercount.

PR #276, opened 2019-05-05
Add missing ok button on validation warnings

vpdn:
Fixes https://github.com/LoopKit/Loop/issues/946


ps2:
A user reported they have the same issue when max bolus is reached. Could you look for other cases where the refactoring caused a warning w/o an action button?

vpdn:
I followed the references and found two paths that are both already fixed in the PR. The warnings affected are:

1. `"The maximum absorption time is ..."`
2. `"The maximum allowed amount is ... grams"`

If there's another related issue filed, I could also have a look. Do you have an issue number?

PR #289, opened 2019-07-31
Fix bug that was incorrectly extending end date of old override

ps2:
The sequence of:
1. Enable an overrride
2. Cancel the override early
3. Enable another override

...was actually extending the end date of the original override to the moment before the next override was enabled. The piece of logic was supposed to truncate an enabled override if another override was enabled on top of it, but it wasn't checking the actual end date of the override to verify it needed to be adjusted.

The result is that on (3), the historical CR and ISF schedules were being changed, which meant that the effects of doses during that time became wildly different, which in turn would throw off ICE and hence carb absorption.

Fixes https://github.com/LoopKit/Loop/issues/1030, https://github.com/LoopKit/Loop/issues/1006, and https://github.com/LoopKit/Loop/issues/1043


Kdisimone:
Massive bug fix. Thank you!

amdee1:
Wow! Need to test this now. Thank you!

mylma:
Wow! Need to test this. Thank you!

PR #322, opened 2020-03-31
Finnish translations update

mylma:
Touchups and adding some previously missing strings.
See my comments here: https://github.com/LoopKit/Loop/pull/1291


ps2:
The project file changes and the `LoopKit/fi.lproj/InfoPlist.strings` file is not needed; they appear to be localizing some Info.plist values that I think are already in the plist erroneously (they should only need to be in the app's Info.plist).

mylma:
@ps2 OK, done!

PR #323, opened 2020-04-04
Add support for dose logging

novalegra:
This PR adds the ability to log outside insulin doses in Loop through the app itself, in addition to support for having multiple insulin types be active at the same time. Currently, when an outside dose is logged in HealthKit, Loop assumes that it is the same type of insulin as is currently present in the pump, which is not always the case. This change allows the user to select the insulin model if the dose is being logged; otherwise, if a bolus, basal, or temp basal is given via the pump, the current dose model is saved as a part of the dose entry. Even if the user doesn't log doses, adding insulin model information allows for correct accounting of IOB and insulin effects if the insulin type in the pump is changed (for example, from Humalog to Fiasp). The insulin model of the dose is persisted in the Loop's data store and in HealthKit, and is non-user-editable.

Loop/LoopKit tests have been updated accordingly based on the changes made, and additional tests were added with multiple insulin types active at once. Happy to add more tests if there's anything I missed.

While this PR does not add additional insulin curves, the underlying changes to the dose model could support new insulin types if they were added into Loop (such as Afrezza). A proposed Afrezza model is included in the code, but is disabled from a UI standpoint, pending additional testing; I can remove the Afrezza model if that would avoid confusion!

Current Testing Status:
- Has been tested with Dexcom G5 CGM, the mock pump, and Omnipod.
- Needs to be tested with Medtronic pumps.
This PR involves changes to both Loop & LoopKit; the Loop PR (https://github.com/LoopKit/Loop/pull/1296) has the same title.

Screen Shot 2020-04-17 at 8 24 02 PM


LoopKit/InsulinKit/PumpEvent+CoreDataClass.swift
```diff
@@ -162,6 +234,18 @@ extension PumpEvent {
...
+            // ANNA TODO: could this be more elegant?
+            if let model = entry.insulinModel as? ExponentialInsulinModel {
+                modelPeak = model.peakActivityTime
+                modelType = .exponential
+            } else if let model = entry.insulinModel as? ExponentialInsulinModelPreset {
+                modelPeak = (model.getExponentialModel() as! ExponentialInsulinModel).peakActivityTime
+                modelType = .exponential
+            } else if let _ = entry.insulinModel as? WalshInsulinModel {
+                modelType = .walsh
+            }
```
novalegra: Not sure if there's a better way to assign (or structure) the persisted insulin model data. I currently have 4 fields that are used to recreate the model: the type of model (exponential, walsh, or none), the peak time (if exponential), the DIA, and the delay.

LoopKit/InsulinKit/DoseEntry.swift
```diff
@@ -18,7 +18,13 @@ public struct DoseEntry: TimelineValue, Equatable {
     public let unit: DoseUnit
     public let deliveredUnits: Double?
     public let description: String?
+    public let insulinModel: InsulinModel?
     internal(set) public var syncIdentifier: String?
+    
+    // ANNA TODO: this is needed due to the fact the insulin model doesn't conform to equatable; would it be possible to avoid having to specify this?
+    public static func == (lhs: DoseEntry, rhs: DoseEntry) -> Bool {
+        return lhs.type == rhs.type && lhs.startDate == rhs.startDate && lhs.endDate == rhs.endDate && lhs.value == rhs.value && lhs.deliveredUnits == rhs.deliveredUnits && lhs.description == rhs.description && lhs.insulinModel?.effectDuration == rhs.insulinModel?.effectDuration && lhs.insulinModel?.delay == rhs.insulinModel?.delay
+    }
```
novalegra: Is there a way to make InsulinModel conform to equatable so this override isn't needed?

LoopKit/InsulinKit/DoseStore.swift
```diff
@@ -1269,7 +1289,14 @@ extension DoseStore {
                     }
                     return dose.trimmed(to: basalDosingEnd)
                 }
-                let glucoseEffects = trimmedDoses.glucoseEffects(insulinModel: insulinModel, insulinSensitivity: insulinSensitivitySchedule)
+                // Update the longest effect duration based on the doses retrieved
+                // ANNA TODO: any way to make this more "clean"? Essensially, the aim is to find the longest effect duration among the doses
+                self.longestEffectDuration = trimmedDoses.max {
+                    ($0.insulinModel?.effectDuration ?? defaultInsulinModel.effectDuration)
+                    < ($1.insulinModel?.effectDuration ?? defaultInsulinModel.effectDuration)}?.insulinModel?.effectDuration
+                    ?? self.longestEffectDuration
```
novalegra: Is there any way to make this more "clean"? Essentially, this line is finding the longest effect duration among the doses, defaulting back on the current longestEffectDuration if the doses don't have any associated insulin models.

LoopKit/InsulinKit/ExponentialInsulinModelPreset.swift
```diff
@@ -0,0 +1,87 @@
+//
```
novalegra, line 1: This file was previously located in Loop. It's the same except for the Afrezza model being added.

LoopKit/InsulinKit/HKQuantitySample+InsulinKit.swift
```diff
@@ -57,6 +73,20 @@ extension HKQuantitySample {
...
+        metadata[MetadataKeyInsulinCurveDelay] = dose.insulinModel?.delay
+
+        if let model = dose.insulinModel as? ExponentialInsulinModel {
+            metadata[MetadataKeyInsulinCurvePeak] = model.peakActivityTime
+            metadata[MetadataKeyInsulinCurveType] = InsulinModelType.exponential.rawValue
+        } else if let model = dose.insulinModel as? ExponentialInsulinModelPreset {
+            metadata[MetadataKeyInsulinCurvePeak] = (model.getExponentialModel() as! ExponentialInsulinModel).peakActivityTime
+           metadata[MetadataKeyInsulinCurveType] = InsulinModelType.exponential.rawValue
+        } else if let _ = dose.insulinModel as? WalshInsulinModel {
+            metadata[MetadataKeyInsulinCurveType] = InsulinModelType.walsh.rawValue
```
novalegra: This is another place where I'm wondering if there's a more elegant way to assign the values.

LoopKit/InsulinKit/HKQuantitySample+InsulinKit.swift
```diff
@@ -92,6 +122,24 @@ extension HKQuantitySample {
...
+        switch modelType {
+        case .walsh:
+            model = WalshInsulinModel(actionDuration: metadata?[MetadataKeyInsulinCurveDuration] as! TimeInterval, delay: (metadata?[MetadataKeyInsulinCurveDelay] ?? 600) as! TimeInterval)
+        case .exponential:
+            model = ExponentialInsulinModel(actionDuration: metadata?[MetadataKeyInsulinCurveDuration] as! TimeInterval, peakActivityTime: metadata?[MetadataKeyInsulinCurvePeak] as! TimeInterval, delay: (metadata?[MetadataKeyInsulinCurveDelay] ?? 600.0) as! TimeInterval)
+        default:
+            break
+        }
+        
+        return model
```
novalegra: I'm guessing there's a shorter way to do this that I'm not thinking about - is it possible to return from the switch statement directly?

LoopKit/Persistence/PersistenceController.swift
```diff
@@ -177,7 +177,7 @@ public final class PersistenceController {
                 }
             }
 
-            let storeURL = directoryURL.appendingPathComponent("Model.sqlite")
+            let storeURL = directoryURL.appendingPathComponent("Model4.sqlite")
```
novalegra: I created multiple models as I went through the development process; I'm guessing it would be appropriate to get rid of models 2 & 3, and have the jump be from model.sqlite -> model4.sqlite?

LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
```diff
@@ -84,6 +84,61 @@ class CachedInsulinDeliveryObject: NSManagedObject {
...
+            willAccessValue(forKey: "modelDelay")
+            defer { didAccessValue(forKey: "modelDelay") }
+            return primitiveModelDelay?.doubleValue
+        }
+        set {
+            willChangeValue(forKey: "modelDelay")
+            defer { didChangeValue(forKey: "modelDelay") }
+            primitiveModelDelay = newValue != nil ? NSNumber(value: newValue!) : nil
+        }
+    }
```
mpangburn: I'm wondering if—rather than describing an exponential model by the parameters we require in the generic case—we could simply store a representation of the preset model the user selects (in which case you'd still need `modelDuration` for Walsh, but you could remove these two parameters).

There's generally a funkiness in that InsulinModel is a protocol, but the actual set of insulin models we offer is a fixed enumeration, InsulinModelSettings (which it might make sense to move from LoopCore to LoopKit). I think we should use the structure of the latter when serializing for storage.

I should note that Tidepool is also looking into this general area, since it's necessary to serialize the user's insulin model in some way for sending to the backend.

LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
```diff
@@ -94,6 +149,35 @@ class CachedInsulinDeliveryObject: NSManagedObject {
 
 
 extension CachedInsulinDeliveryObject {
+    var insulinModel: InsulinModel? {
+        get {
+            switch modelType {
+            case .walsh:
+                return WalshInsulinModel(actionDuration: modelDuration!, delay: modelDelay ?? 600)
+            case .exponential:
+                return ExponentialInsulinModel(actionDuration: modelDuration!, peakActivityTime: modelPeak!, delay: modelDelay ?? 600)
```
mpangburn: In these situations, although it's designed such that e.g. `modelDuration` should never be `nil` when `modelType` is `walsh`, it's safer to unwrap these values with `guard let` and return `nil` if they aren't present.

We wouldn't expect it to happen in practice, but if something goes awry, better to lose the insulin model than trap on access, I think.

LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
```diff
@@ -94,6 +149,35 @@ class CachedInsulinDeliveryObject: NSManagedObject {
 
 
 extension CachedInsulinDeliveryObject {
+    var insulinModel: InsulinModel? {
```
mpangburn: Here's an example of where it would be preferable to expose the concrete type InsulinModelSettings rather than the protocol type InsulinModel.

LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
```diff
@@ -94,6 +149,35 @@ class CachedInsulinDeliveryObject: NSManagedObject {
...
+            modelDelay = model.delay
+            if let model = model as? ExponentialInsulinModel {
+                modelPeak = model.peakActivityTime
+                modelType = .exponential
+            } else if let model = model as? ExponentialInsulinModelPreset {
+                modelPeak = (model.getExponentialModel() as! ExponentialInsulinModel).peakActivityTime
+                modelType = .exponential
+            } else if let _ = model as? WalshInsulinModel {
+                modelType = .walsh
+            }
```
mpangburn: And here's a place where you'll see the benefit of using InsulinModelSettings: you'll be able to exhaustively switch over `model`, rather than casting dynamically; the latter is error-prone if a new insulin model were to be added.

LoopKit/InsulinKit/DoseEntry.swift
mpangburn: Here's another place where you'll see the gain of using InsulinModelSettings; you should be able to adopt Equatable more naturally.

LoopKit/InsulinKit/DoseEntry.swift
```diff
@@ -41,6 +47,7 @@ public struct DoseEntry: TimelineValue, Equatable {
         self.description = description
         self.syncIdentifier = syncIdentifier
         self.scheduledBasalRate = scheduledBasalRate
+        self.insulinModel = insulinModel // if nil, it's assumed that the insulin model is the type of insulin the pump dispenses
```
mpangburn: Might be worth putting this comment on the `init` declaration itself for better visibility to the callee.

LoopKit/InsulinKit/DoseStore.swift
```diff
@@ -696,11 +700,14 @@ extension DoseStore {
 
      - parameter events: An array of new pump events. Pump events should have end times reflective of when delivery is actually expected to be finished, as doses that end prior to a reservoir reading are ignored when reservoir data is being used.
      - parameter lastReconciliation: The date that pump events were most recently reconciled against recorded pump history. Pump events are assumed to be reflective of delivery up until this point in time. If reservoir values are recorded after this time, they may be used to supplement event based delivery.
+     - parameter updateLastReconciliation: Whether or not the last reconciliation time should be updated with the new value.
```
mpangburn: Not having finished reviewing, it's not immediately clear to me the circumstances under which this parameter should be `false`.

LoopKit/InsulinKit/DoseStore.swift
mpangburn: You can clear out the `nil` values then grab the maximum:
```swift
trimmedDoses.compactMap({ $0.insulinModel?.effectDuration }).max() ?? self.longestEffectDuration
```

LoopKit/InsulinKit/ExponentialInsulinModel.swift
```diff
@@ -72,6 +72,12 @@ extension ExponentialInsulinModel: CustomDebugStringConvertible {
     }
 }
 
+extension ExponentialInsulinModel: Equatable {
+    public static func ==(lhs: ExponentialInsulinModel, rhs: ExponentialInsulinModel) -> Bool {
+        return abs(lhs.actionDuration - rhs.actionDuration) < .ulpOfOne && abs(lhs.peakActivityTime - rhs.peakActivityTime) < .ulpOfOne && abs(lhs.delay - rhs.delay) < .ulpOfOne
+    }
+}
```
mpangburn: You've done well to follow precedent here in comparing to `ulpOfOne`, though in practice `ulpOfOne` isn't a great epsilon. In general, two ExponentialInsulinModel instances will either be created with the exact same values, or very different ones, and the values are static. You can safely adopt the auto-generated Equatable conformance here.

LoopKit/InsulinKit/PumpEvent+CoreDataClass.swift
mpangburn: Same comments apply here re moving to using InsulinModelSettings as the representation.

mpangburn (changes requested):
Nice! Switching the primary model representation from InsulinModel to InsulinModelSettings will help simplify several areas. Otherwise, just a few suggestions.

LoopKit/InsulinKit/ExponentialInsulinModelPreset.swift
novalegra: Afrezza model was removed

novalegra:
@mpangburn I've made all of your suggested changes! The two biggest were the migration to InsulinModelSettings (which minimizes the amount of information that has to be store in CoreData/HealthKit) and creating a new persisted type for logged doses (which includes UI changes for the Insulin Delivery screen). 
Screen Shot 2020-04-17 at 8 21 27 PM

novalegra:
@ps2 Let me know when the Tidepool dose store changes have landed in DIY, and I can update my PRs accordingly

PR #327, opened 2020-04-20
Fix Temp Override missing translations issue

mylma:
This PR fixes the missing translations issue by replacing the `NSLocalizedStrings` with `LocalizedStrings`.

https://github.com/LoopKit/Loop/issues/1304


mpangburn:
Thanks @mylma !

PR #329, opened 2020-05-08
Add optional batterylevel to cgmmanagers

dabear:
See Loop PR here: https://github.com/LoopKit/Loop/pull/1325

### DIFF
--- a/LoopKit/DeviceManager/CGMManager.swift
+++ b/LoopKit/DeviceManager/CGMManager.swift
@@ -51,8 +51,8 @@ public protocol CGMManager: DeviceManager {
 
     var appURL: URL? { get }
 
-    // The cgmManagers associated batteryLevel, from 1 to 100
-    var batteryLevel : Int? { get }
+    // The cgmManagers associated batteryLevel, from 0 to 1
+    var batteryLevel : Double? { get }
 
     /// Whether the device is capable of waking the app
     var providesBLEHeartbeat: Bool { get }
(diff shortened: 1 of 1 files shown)


LoopKit/DeviceManager/CGMManager.swift
```diff
@@ -51,6 +51,9 @@ public protocol CGMManager: DeviceManager {
 
     var appURL: URL? { get }
 
+    // The cgmManagers associated batteryLevel, from 1 to 100
+    var batteryLevel : Int? { get }
```
ps2: In Loop, we typically represent percentages as doubles ranging from 0-1

ps2 (changes requested):

dabear:
Concerns from ps2 fixed and verified

dabear:
@ps2 let me know if I should do anything more with this to get it merged

LoopKit/DeviceManager/CGMManager.swift
```diff
@@ -71,6 +74,12 @@ public protocol CGMManager: DeviceManager {
     func fetchNewDataIfNeeded(_ completion: @escaping (CGMResult) -> Void) -> Void
 }
 
+public extension CGMManager {
```
ps2, line 77: We generally try to avoid default implementations for protocols.  See https://medium.com/@xcadaverx/default-protocol-extension-caveats-88c6ece27fdb 

I think this should be removed, which means that implementations will need to be added for the existing CGM managers.

ps2 (changes requested):

dabear:
Will rework this and possibly resubmit

PR #335, opened 2020-08-15
Add dose logging infrastructure + UI

novalegra:
## Description
This PR adds the ability to log outside insulin doses in Loop through the app itself, in addition to support for having multiple insulin types be active at the same time. Currently, when an outside dose is logged in HealthKit, Loop assumes that it is the same type of insulin as is currently present in the pump, which is not always the case. This change allows the user to select the insulin model if the dose is being logged; otherwise, if a bolus, basal, or temp basal is given via the pump, the *current dose model* is saved as a part of the dose entry. Even if the user doesn't log doses, adding insulin model information allows for correct accounting of IOB and insulin effects if the insulin type in the pump is changed (for example, from Humalog to Fiasp). The insulin model of the dose is persisted in the Loop's data store and in HealthKit, and is not editable.

Tests have been updated accordingly based on the changes made, and additional tests were added with multiple insulin types active at once.

While this PR does not add additional insulin curves, the underlying changes to the dose model could support new insulin types if they were added into Loop (such as Afrezza).

## Current Testing Status:
- Has been tested with Dexcom G5 CGM, the mock pump, and Omnipod.
- Needs to be tested with Medtronic pumps.

## Other Relevant Info
- This PR also involves changes to Loop, see https://github.com/LoopKit/Loop/pull/1357
- The infrastructure changes are the same as in https://github.com/LoopKit/LoopKit/pull/323 (which was reviewed by Michael), though some of the UIs have changed so they're in SwiftUI

Screen Shot 2020-08-20 at 4 42 24 PM
Screen Shot 2020-08-20 at 4 42 15 PM
Screen Shot 2020-08-20 at 4 42 03 PM
Screen Shot 2020-08-20 at 4 41 08 PM


LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
```diff
@@ -84,6 +84,35 @@ class CachedInsulinDeliveryObject: NSManagedObject {
             primitiveProgrammedTempBasalRate = NSNumber(value: rate)
         }
     }
+    
+    var modelDuration: Double? {
```
novalegra: In Core Data, insulin model info is stored as a duration (to add support for Walsh, if desired) and a model type (Walsh, exponential adult, exponential child, or Fiasp)

LoopKit/InsulinKit/InsulinMath.swift
```diff
@@ -301,7 +325,8 @@ extension DoseEntry {
         }
         
         let netTempBasalDoses = self.annotated(with: basalRateSchedule)
-        return netTempBasalDoses.glucoseEffects(insulinModel: insulinModel, insulinSensitivity: insulinSensitivity)
+        // Assume that temp basals will be delivered by the pump
+        return netTempBasalDoses.glucoseEffects(defaultModel: insulinModel, longestEffectDuration: insulinModel.effectDuration, insulinSensitivity: insulinSensitivity)
```
novalegra: Keeping track of "longestEffectDuration" ensures we include all the doses that could still be active in the glucose effects

LoopKit/InsulinKit/OutsideDoseEvent.swift
```diff
@@ -0,0 +1,15 @@
...
+//  OutsideDoseEvent.swift
+//  LoopKit
+//
+//  Created by Anna Quinlan on 4/17/20.
+//  Copyright © 2020 LoopKit Authors. All rights reserved.
+//
+import Foundation
+import CoreData
+
+class OutsideDoseEvent: PumpEvent {
```
novalegra: An outside dose event contains basically all the info of a `PumpEvent`, but makes it so we can easily distinguish between a logged dose and a pump-delivered dose

LoopKit.xcodeproj/project.pbxproj
```diff
@@ -749,7 +748,14 @@
...
 		E96DCB5A24AF74AC007117BC /* SuspendThresholdEditor.swift in Sources */ = {isa = PBXBuildFile; fileRef = E96DCB5924AF74AC007117BC /* SuspendThresholdEditor.swift */; };
 		E96DCB5E24AF7DC7007117BC /* BasalRatesInformationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E96DCB5D24AF7DC7007117BC /* BasalRatesInformationView.swift */; };
+		E9CDA72724E3A4BA0049B0D5 /* OutsideDoseEvent.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9CDA72624E3A4BA0049B0D5 /* OutsideDoseEvent.swift */; };
+		E9CDA72824E3A4BA0049B0D5 /* OutsideDoseEvent.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9CDA72624E3A4BA0049B0D5 /* OutsideDoseEvent.swift */; };
+		E9CDA72A24E3A4F60049B0D5 /* PersistedOutsideDoseEvent.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9CDA72924E3A4F60049B0D5 /* PersistedOutsideDoseEvent.swift */; };
+		E9CDA72B24E3A4F60049B0D5 /* PersistedOutsideDoseEvent.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9CDA72924E3A4F60049B0D5 /* PersistedOutsideDoseEvent.swift */; };
```
novalegra: I have no clue why these are showing up as doubles, but the build does fail when I try to manually remove them

LoopKit/Persistence/Model.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>Modelv2.xcdatamodel</string>
+	<string>Modelv3.xcdatamodel</string>
```
novalegra: There's a new model because we added 2 new properties to CoreData

LoopKitTests/InsulinMathTests.swift
```diff
@@ -33,6 +33,10 @@ class InsulinMathTests: XCTestCase {
     var fixtureDateformatter: DateFormatter!
     
     private let fixtureTimeZone = TimeZone(secondsFromGMT: -0 * 60 * 60)!
+    
+    private let model = WalshInsulinModel(actionDuration: TimeInterval(hours: 4))
+    
+    private let exponentialModel = ExponentialInsulinModel(actionDuration: TimeInterval(minutes: 360), peakActivityTime: TimeInterval(minutes: 75))
```
novalegra: Pull the insulin models out of the individual tests to make it cleaner

LoopKitTests/InsulinMathTests.swift
```diff
@@ -93,7 +97,7 @@ class InsulinMathTests: XCTestCase {
         }
     }
 
-    func loadDoseFixture(_ resourceName: String) -> [DoseEntry] {
+    func loadDoseFixture(_ resourceName: String, insulinModel: InsulinModel? = nil) -> [DoseEntry] {
```
novalegra: `insulinModel` was added so the doses have the intended model

LoopKitTests/InsulinMathTests.swift
```diff
@@ -346,17 +344,55 @@ class InsulinMathTests: XCTestCase {
...
+            DoseEntry(type: .tempBasal, startDate: f("2018-05-15 14:42:36 +0000"), endDate: f("2018-05-15 14:42:36 +0000"), value: 0.0, unit: .unitsPerHour, syncIdentifier: "1600646a074f12", scheduledBasalRate: nil),
+            DoseEntry(type: .tempBasal, startDate: f("2018-05-15 14:32:51 +0000"), endDate: f("2018-05-15 15:02:51 +0000"), value: 1.8999999999999999, unit: .unitsPerHour, syncIdentifier: "16017360074f12", scheduledBasalRate: nil),
+            DoseEntry(type: .bolus, startDate: f("2018-05-15 14:52:51 +0000"), endDate: f("2018-05-15 15:52:51 +0000"), value: 0.9, unit: .units, syncIdentifier: "01004a004a006d006e22354312", scheduledBasalRate: nil, insulinModelSetting: InsulinModelSettings(model: child)),
+        ]
+
+        let iobWithModel = dosesWithModel.insulinOnBoard(defaultModel: exponentialModel, longestEffectDuration: exponentialModel.effectDuration)
+
+        XCTAssertEqual(iobWithoutModel.count, iobWithModel.count)
```
novalegra, line 383: Checking that adding the model doesn't change the length of the prediction

LoopKitTests/InsulinMathTests.swift
```diff
@@ -346,17 +344,55 @@ class InsulinMathTests: XCTestCase {
...
+            DoseEntry(type: .bolus, startDate: f("2018-05-15 14:44:46 +0000"), endDate: f("2018-05-15 14:44:46 +0000"), value: 0.9, unit: .units, syncIdentifier: "01004a004a006d006e22354312", scheduledBasalRate: nil),
+            DoseEntry(type: .tempBasal, startDate: f("2018-05-15 14:42:36 +0000"), endDate: f("2018-05-15 14:42:36 +0000"), value: 0.0, unit: .unitsPerHour, syncIdentifier: "1600646a074f12", scheduledBasalRate: nil),
+            DoseEntry(type: .tempBasal, startDate: f("2018-05-15 14:32:51 +0000"), endDate: f("2018-05-15 15:02:51 +0000"), value: 1.8999999999999999, unit: .unitsPerHour, syncIdentifier: "16017360074f12", scheduledBasalRate: nil),
+            DoseEntry(type: .bolus, startDate: f("2018-05-15 14:52:51 +0000"), endDate: f("2018-05-15 15:52:51 +0000"), value: 0.9, unit: .units, syncIdentifier: "01004a004a006d006e22354312", scheduledBasalRate: nil),
+        ]
+
+        let iobWithoutModel = doses.insulinOnBoard(defaultModel: exponentialModel, longestEffectDuration: exponentialModel.effectDuration)
+
+        let dosesWithModel = [
```
novalegra, line 372: Unit test that IOB accounting is correct when there doses with different insulin model curves

LoopKitUI/InsulinKit/ErrorBackgroundView.swift
```diff
@@ -8,8 +8,15 @@
 
 import UIKit
 
-class ErrorBackgroundView: UIView {
+public class ErrorBackgroundView: UIView {
```
novalegra, line 11: Needed so `InsulinDeliveryTableViewController` in Loop is able to set the label

LoopKitUI/Views/ExpandablePicker.swift
```diff
@@ -0,0 +1,63 @@
...
+extension Binding {
+    func onChange(_ handler: @escaping (Value) -> Void) -> Binding<Value> {
+        return Binding(
+            get: { self.wrappedValue },
+            set: { selection in
+                self.wrappedValue = selection
+                handler(selection)
+        })
+    }
+}
```
novalegra: Make it so we can update values in the view model (this functionality will be built-in in iOS 14, but isn't available in iOS 13)

LoopKitUI/Views/ExpandableDatePicker.swift
```diff
@@ -0,0 +1,50 @@
...
+        self.onUpdate = onUpdate
+        self.date = date
+    }
+    
+    public var body: some View {
+        ZStack(alignment: .topLeading) {
+            // ANNA TOOD: fix buggy animations
+            DatePicker(
+                "",
+                selection: $date.onChange(onUpdate),
```
novalegra: This is a little bit hacky, but since bindings can't be passed into a child view and then into a date picker, this is the best solution I was able to come up with

LoopKitUI/Views/ExpandablePicker.swift
```diff
@@ -0,0 +1,66 @@
...
+                Text(items[pickerIndex])
+                .foregroundColor(.gray)
+            }
+            .padding(.vertical, 5)
+            .frame(minWidth: 0, maxWidth: .infinity).onTapGesture {
+                self.pickerShouldExpand.toggle()
+            }
+            if pickerShouldExpand {
+                HStack {
+                    Picker(selection: $pickerIndex.onChange(onUpdate), label: Text("")) {
```
novalegra: This is a little bit hacky, but since bindings can't be passed into a child view and then into a picker, this is the best solution I was able to come up with

LoopKitUI/Views/ExpandableDatePicker.swift
```diff
@@ -0,0 +1,50 @@
...
+        let today = Date()
+        self.pickerRange = pickerRange ?? today.addingTimeInterval(-.hours(6))...today.addingTimeInterval(.hours(6))
+        self.text = text
+        self.onUpdate = onUpdate
+        self.date = date
+    }
+    
+    public var body: some View {
+        ZStack(alignment: .topLeading) {
+            // ANNA TOOD: fix buggy animations
```
novalegra: The animations for the date picker appear a bit weirdly (though not for the other picker for some reason...). I can't find a way to disable the animation (setting `.animation(nil` doesn't do anything), but if you have ideas I'd love to find a fix

novalegra:
@ps2 I think this is ready for review now. I'm still not happy with the `DatePicker` animations (there's a more complex hack I could try depending on the timeline for this being merged), but I think otherwise everything is in there

ps2:
Can this be rebased onto dev?

novalegra:
> Can this be rebased onto dev?

Done for Loop & LoopKit. Had to make a few changes due to the refactoring of `BolusViewModel`, will test later today

novalegra:
> Done for Loop & LoopKit. Had to make a few changes due to the refactoring of `BolusViewModel`, will test later today

Tested and couldn't find any bugs introduced by the changes I made to support the refactoring.

ps2:
Ok, we're in sync with Tidepool ATM, and should have a window. If you can resolve the conflicts here, I'll prioritize it for landing in dev.

LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
ps2: I think we want to be storing the Insulin *type* in HK, and DoseStore. I.e. Fast-Acting, Fiasp, Afrezza, etc. Not specific model of insulin activity. The mapping of insulin type to activity curve would vary based on insulin type and user preferences. I.e. fiasp would map to fiasp curve, and Fast Acting could possibly map to one of the adult or child exponential models, depending on what the user prefers as a Loop setting.

LoopKit/InsulinKit/PersistedOutsideDoseEvent.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  PersistedOutsideDoseEvent.swift
+//  LoopKit
+//
+//  Created by Anna Quinlan on 4/17/20.
+//  Copyright © 2020 LoopKit Authors. All rights reserved.
+//
+import Foundation
+
+public struct PersistedOutsideDoseEvent {
```
ps2: Would prefer that we use existing types (i.e. primarily CachedInsulinDeliveryObject) to track outside doses. Is there some reason that this doesn't work?

ps2 (changes requested):
In general looks good, but I do have some comments about how we're modeling these, that might result in significant changes.

LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
```diff
@@ -82,6 +82,35 @@ class CachedInsulinDeliveryObject: NSManagedObject {
             primitiveProgrammedTempBasalRate = NSNumber(value: rate)
         }
     }
+
+    var insulinModelCategory: InsulinModelCategory? {
```
ps2: We aren't storing a category of insulin model in HK, but rather a field that represents the *type* of insulin. The insulin *model* that Loop uses may be related to the type of insulin, and loop may switch models based on the type of insulin. But I don't think we should call this a `insulinModelCategory`.  Something like `insulinType` would be better here.

ps2 (changes requested):
I think there is just a renaming left to be done. Normally a rename wouldn't be a blocker, but I think this one is semantically pretty important, and because it involves names of keys in HK and columns in core data.

LoopKit/InsulinKit/CachedInsulinDeliveryObject+CoreDataClass.swift
novalegra: Thanks so much for the suggestion - I was struggling to think of a better name for it!

PR #336, opened 2020-08-22
Add way to view override history in Loop

novalegra:
## Description
Currently, there's no great way to view the override history in Loop; the issue report does contain the history, but it's not in a very human-readable format. This PR adds an override history screen and some under-the-hood improvements to override accounting in order to correctly determine the end date of an override. The history screen is written in SwiftUI, which bumps the target version to iOS 13. I know the spacing on the elements in the list isn't ideal, but there really isn't any way to change the spacing in SwiftUI. One other thing to note is that the history is filtered so it doesn't include overrides that were scheduled but never activated. 

Since I started this before the Tidepool merge (which also bumps the target to iOS 13), I left the targets in the project file in this PR; they can easily be removed if the Tidepool stuff lands first. There's a teeny associated Loop PR (https://github.com/LoopKit/Loop/pull/1363)
## Demo Video
[](http://www.youtube.com/watch?v=BFFl4rHE1bg "Demo")


LoopKit/TemporaryScheduleOverride.swift
```diff
@@ -50,14 +50,27 @@ public struct TemporaryScheduleOverride: Hashable {
...
+    }
 
     public var duration: Duration {
         didSet {
             precondition(duration.timeInterval > 0)
         }
     }
 
-    public var endDate: Date {
+    public var scheduledEndDate: Date {
```
novalegra, line 73: I changed this to be `scheduledEndDate` to more accurately reflect what type of "end" it is

novalegra: I tried using this for duration on the history screens, but it wasn't being updated if the override ended early

LoopKit/DailyQuantitySchedule+Override.swift
```diff
@@ -16,7 +16,7 @@ extension GlucoseRangeSchedule {
         }
 
         let doubleRange = targetRange.doubleRange(for: unit)
-        let rangeOverride = GlucoseRangeSchedule.Override(start: override.startDate, end: override.endDate, value: doubleRange)
+        let rangeOverride = GlucoseRangeSchedule.Override(start: override.startDate, end: override.scheduledEndDate, value: doubleRange)
```
novalegra: This is just a variable rename

LoopKit/TemporaryScheduleOverrideHistory.swift
```diff
@@ -8,16 +8,15 @@
...
 
 private struct OverrideEvent: Equatable {
-    enum End: Equatable {
-        case natural
-        case early(Date)
-        case deleted // Ended before started
-    }
 
     var override: TemporaryScheduleOverride
-    var end: End = .natural
```
novalegra, line 20: Move the `end` to each `TemporaryScheduleOverride` so it's accessible for end-date-calculation

LoopKit/TemporaryScheduleOverride.swift
```diff
@@ -50,14 +50,27 @@ public struct TemporaryScheduleOverride: Hashable {
...
+    
+    public var actualEndDate: Date {
+        switch actualEnd {
+        case .natural:
+            return scheduledEndDate
+        case .early(let endDate):
+            return endDate
+        case .deleted:
+            return scheduledEndDate
+        }
```
novalegra, line 64: This logic is from `TemporaryScheduleOverrideHistory`

LoopKit/TemporaryScheduleOverrideHistory.swift
```diff
@@ -163,6 +162,11 @@ public final class TemporaryScheduleOverrideHistory {
             base.applyingCarbRatioMultiplier(from: override, relativeTo: referenceDate)
         }
     }
+    
+    public func getEvents(relativeTo referenceDate: Date = Date()) -> [TemporaryScheduleOverride] {
+        filterRecentEvents(relativeTo: referenceDate)
+        return recentEvents.map { $0.override }
+    }
```
novalegra, line 202: For getting events for the history

LoopKitUI/Views/OverrideHistoryCollectionViewCell.swift
```diff
@@ -0,0 +1,79 @@
...
+//  LoopKitUI
+//
+//  Created by Anna Quinlan on 8/6/20.
+//  Copyright © 2020 LoopKit Authors. All rights reserved.
+//
+
+import UIKit
+
+
+final class OverrideHistoryCollectionViewCell: UICollectionViewCell, IdentifiableClass {
```
novalegra, line 12: This is basically the same as the `CustomOverrideCollectionViewCell`, but with some storyboard adj to change the icon and title.

novalegra:
Closing because I incorporated this into my Siri overrides PR

PR #343, opened 2020-10-22
Add Siri support for overrides

novalegra:
Currently, overrides aren't visible to Siri (meaning they can't be set with Shortcuts or via voice). There also isn't a great way to view the override history in Loop; the issue report does contain the history, but it's not in a very human-readable format. 

This PR:
1) has some minor LoopKit-level tweaks so Loop can donate an override preset intent when one is set
3) adds an override history screen
4) makes some under-the-hood improvements to override accounting in order to correctly determine the end date of an override

[](http://www.youtube.com/watch?v=ufLSOx6KbX8 "Demo video")


LoopKit/TemporaryScheduleOverride.swift
```diff
@@ -50,14 +50,27 @@ public struct TemporaryScheduleOverride: Hashable {
     public var startDate: Date
     public let enactTrigger: EnactTrigger
     public let syncIdentifier: UUID
+    
+    public var actualEnd: End = .natural
```
novalegra, line 54: I moved this logic out of TemporayScheduleOverrideHistory and into here so the history screens can have access to it

LoopKitUI/UIColor.swift
```diff
@@ -15,6 +15,10 @@ private func BundleColor(_ name: String, compatibleWith traitCollection: UITrait
 }
 
 extension UIColor {
+    @nonobjc static let lightenedInsulin = BundleColor("Lightened Insulin") ?? systemOrange
+    
+    @nonobjc static let darkenedInsulin = BundleColor("Darkened Insulin") ?? systemOrange
+
```
novalegra, line 21: Colors for override percentage gauge bar

ps2: Colors used in LoopKit should be provided by the app layer in some way. Passing colors into the view constructor, or assigning attributes on the view, or setting up Environment attributes for SwiftUI views.

LoopKit/TemporaryScheduleOverride.swift
ps2: Can OverrideEvent be cleaned up to not need this then, since OverrideEvent contains a TemporaryScheduleOverride?

ps2 (commented):
Generally looks good code wise; colors should not be pulling from the LoopKit bundle, though. I haven't run it yet, but just looked at the code.

LoopKit/TemporaryScheduleOverride.swift
novalegra: Yes & done!

PR #344, opened 2020-10-23
Only fetch reservoir values after reconciled events

elnjensen:
Rather than using reservoir values after the end of the last basal, only use them after the end of the reconciled events, to avoid double-counting of boluses.

This addresses the double-counting issue for Medtronic pumps, raised in Loop as issues LoopKit/Loop#845, LoopKit/Loop#1008, LoopKit/Loop#1047, LoopKit/Loop#1373.

Some additional discussion and testing results are in [this thread](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Possible.20solution.2Fclue.20for.20double-counted.20boluses) in Zulip.


ps2:
I think the clock sync should cover us missing too much insulin in the case of clock drift and failure to read history. I'm going to merge this to dev (which still needs a good amount of testing). Thanks Eric!

PR #353, opened 2021-02-02
Partial fix scroll issue glucose range and fractional quantity pickers

Lytrix:
A first attempt to fix the [scroll animation issue when 2 pickers are used](https://github.com/LoopKit/Loop/issues/1479).

It seems some race condition is happening, causing the scroll animation to run on the other picker after changing the value.

It might be caused by the forced update to  ensure the selectable picker values update when either bound changes.
Creating the same indexes on both pickers within the function.

As a test I disabled them. It seems to solve the issue at least on the FractionalQuantityPicker.swift file, but on the GlucoseRangePicker.swift the upperrange still gets an unintended animation, while the lower range does seem to work properly.

It might be needed to create unique variables for the id's or create a state function outside of the Picker to ensure the updates happen in the right sequential order, but I don't know how to set that up in a correct way. I have only done it for a website once in Vue.js.

### DIFF
--- a/LoopKitUI/Views/FractionalQuantityPicker.swift
+++ b/LoopKitUI/Views/FractionalQuantityPicker.swift
@@ -123,7 +123,7 @@ public struct FractionalQuantityPicker: View {
                 colorForValue: colorForWhole
             )
             // Ensure whole picker color updates when fraction updates
-            .id(whole + fraction)
+            //.id(whole + fraction)
             .frame(width: availableWidth / 3.5)
             .overlay(
                 Text(separator)
@@ -143,7 +143,7 @@ public struct FractionalQuantityPicker: View {
                 colorForValue: colorForFraction
             )
             // Ensure fractional picker values update when whole value updates
-            .id(whole + fraction)
+            //.id(whole + fraction)
             .frame(width: availableWidth / 3.5)
             .padding(.trailing, spacing + unitLabelWidth)
             .clipped()
--- a/LoopKitUI/Views/GlucoseRangePicker.swift
+++ b/LoopKitUI/Views/GlucoseRangePicker.swift
@@ -100,7 +100,7 @@ public struct GlucoseRangePicker: View {
                 isUnitLabelVisible: false
             )
             // Ensure the selectable picker values update when either bound changes
-            .id(lowerBound...upperBound)
+            //.id(lowerBound...upperBound)
             .frame(width: availableWidth / 3.5)
             .overlay(
                 Text(separator)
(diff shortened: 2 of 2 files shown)


ps2:
Could you reinstate the original ordering and spacing in the code?

ps2:
The comments above the lines you commented indicate the reason they were added; have you tested those scenarios to make sure they still work when commenting them out?

ps2:
I also checked to see if this same approach works for GlucoseRangePicker, but it doesn't appear to.

Lytrix:
@ps2 Sorry for the late update, I did not receive any notifications unfortunately, so did not see the comments. I did continue to further investigate this issue.

> The comments above the lines you commented indicate the reason they were added; have you tested those scenarios to make sure they still work when commenting them out?

Yes, I tested this and the boundaries do update correctly still with these commented out, for example moving to the last value, the other picker updates also the boundary to only that allowed value and vice versa.

> I also checked to see if this same approach works for GlucoseRangePicker, but it doesn't appear to.

Indeed, but it does work for the upper range. When using this approach it does not trigger the lower range animation any more.
Unfortunately it does still animate the upper range value when changing the lower range.
I still haven't found out why unfortunately.

The tests I did so far:
1. The issue is gone when you fixate or use the lowerBoundsRange on the upperrange binding. So maybe the way the upperBoundRange is being updated is causing this, but it seems very similar to the lowerBoundsRange. 
2. Only difference I saw is they way the maxValue is treated differently in comparison with the minValue, but I am not sure this causes it, so I tried some test with with other init values or setting a fixed maxValue, but that did not make any difference.
3. Reversing the order of the GlucoseValuePicker in the HStack also did not make any difference. the lower range value still triggered the animation of the upper range.

Maybe setting an explicit edit state which binding is being edited would be a solution so you could control in which of the glucosepicker an animation is allowed or not here:
https://github.com/LoopKit/LoopKit/blob/dev/LoopKitUI/Views/Settings%20Editors/CorrectionRangeScheduleEditor.swift#L72

Similar like is done on the override range picker to select the preset that is being edited:
https://github.com/LoopKit/LoopKit/blob/dev/LoopKitUI/Views/Settings%20Editors/CorrectionRangeOverridesEditor.swift#L151

It is indeed a tough nut to crack!

ps2:
This approach seems to cause problems with the fractional value updating.  I think a significantly different approach is going to be necessary to actually fix this bug, unfortunately. Maybe using the new `ResizeablePicker` in LoopKit, which is a wrapped UIPickerView.

PR #354, opened 2021-02-08
Fix label for delete button on service account

elnjensen:


ps2:
This was fixed in dev in a different commit. Thanks @elnjensen !

PR #369, opened 2021-08-15

novalegra:
This PR enables outside insulin doses that are logged within Loop to be uploaded to Nightscout. To do this, I updated the Core Data model for `CachedInsulinDeliveryObject` to keep track of when it was changed relative to other changes made on the Core Data managed object context, and added uploading for any `CachedInsulinDeliveryObject`s that don't duplicate data already uploaded in a `PumpEvent`. There is test coverage for these changes and an associated Loop PR (https://github.com/LoopKit/Loop/pull/1522) to update the dose store protocols.


novalegra:
Closing to work more on upload logic

PR #372, opened 2021-08-26

novalegra:
This PR enables outside insulin doses that are logged within Loop to be uploaded to Nightscout. To do this, I updated the Core Data model for CachedInsulinDeliveryObject to keep track of when it was changed relative to other changes made on the Core Data managed object context, and modified the dose upload function to upload CachedInsulinDeliveryObjects (PumpEvents should still be uploaded via the `executePumpEventQuery` function). There is test coverage for these changes.


novalegra:
Closing since my work is now on another branch

PR #378, opened 2021-12-23

novalegra:
This PR enables outside insulin doses that are logged within Loop to be uploaded to Nightscout. To do this, I updated the Core Data model for CachedInsulinDeliveryObject to keep track of when it was changed relative to other changes made on the Core Data managed object context, and modified the dose upload function to upload CachedInsulinDeliveryObjects (PumpEvents will still be uploaded via the `executePumpEventQuery` function). 

There is test coverage for these changes, and I've tested upgrading from dev to this branch (to ensure the Core Data migration is successful).


ps2:
This is working with the latest changes from the Tidepool merge to upload doses from InsulinDeliveryStore.

PR #381, opened 2022-01-27
Restore observer queries startup in HealthKit Stores.

### LINKED ISSUE #1593: `dev` not displaying non-CGM glucose values from Apple Health
I upgraded from 2.2.6 to dev (6e2f25, v2.3.0 (57)) last night, and observed that glucose values recorded to Apple Health from my test kit are no longer appearing in Loop's Glucose display.

I've verified that `observeHealthKitSamplesFromOtherApps` is `true` as shown in an Issue Report.

@marionbarker reported at https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Missing.20test.20kit.20BGs.20on.20dev/near/269142359 that the issue reproduced on their setup as well, and other users have chimed in as well.

I'm on an iPhone 12 Pro running iOS 15.1.1.

whiten:
This restores calls to createQuery() lost in the 8947c0f merge.  At
present, these exist in `master` within HealthKitSampleStore's init, but
were moved into individual stores as part of 2ea6be07, and then removed
later for an unknown reason.

This adds tests for all stores that an observer query has been started
during initialization in the case that authorizationRequired returns
false; if authorization is required, queries will be created as part of
authorize().

Fixes LoopKit/Loop#1593.

### DIFF
--- a/LoopKit/CarbKit/CarbStore.swift
+++ b/LoopKit/CarbKit/CarbStore.swift
@@ -252,7 +252,11 @@ public final class CarbStore: HealthKitSampleStore {
             cacheStore.fetchAnchor(key: CarbStore.healthKitQueryAnchorMetadataKey) { (anchor) in
                 self.queue.async {
                     self.queryAnchor = anchor
-            
+
+                    if !self.authorizationRequired {
+                        self.createQuery()
+                    }
+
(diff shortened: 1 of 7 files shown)


ps2:
Thanks for a great first PR!

PR #382, opened 2022-02-20
Nightscout api plugin dev

gestrich:
See the associated https://github.com/LoopKit/LoopWorkspace/pull/14 for a high level description.

### DIFF
--- a/LoopKitUI/ServiceAuthenticationUI.swift
+++ b/LoopKitUI/ServiceAuthenticationUI.swift
@@ -11,6 +11,8 @@ import LoopKit
 public protocol ServiceAuthenticationUI: ServiceAuthentication {
     // The indexed credentials (e.g. username, password) used to authenticate
     var credentialFormFields: [ServiceCredential] { get }
+    //Form field helper text displayed when completing form
+    var credentialFormFieldHelperMessage: String? { get }
 }
 
 
--- a/LoopKitUI/View Controllers/AuthenticationViewController.swift
+++ b/LoopKitUI/View Controllers/AuthenticationViewController.swift
@@ -62,14 +62,7 @@ public final class AuthenticationViewController<T: ServiceAuthenticationUI>: UIT
     }
 
     var credentials: [(field: ServiceCredential, value: String?)] {
-        switch state {
-        case .authorized:
-
-
-            return authentication.credentials.filter({ !$0.field.isSecret })
-        default:
-            return authentication.credentials
-        }
+        return authentication.credentials
     }
 
     public init(authentication: T) {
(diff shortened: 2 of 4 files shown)


LoopKitUI/ServiceAuthenticationUI.swift
```diff
@@ -11,6 +11,8 @@ import LoopKit
 public protocol ServiceAuthenticationUI: ServiceAuthentication {
     // The indexed credentials (e.g. username, password) used to authenticate
     var credentialFormFields: [ServiceCredential] { get }
```
gestrich, line 13: Support helper text -- Used for Nightscout API Client.

LoopKitUI/View Controllers/AuthenticationViewController.swift
```diff
@@ -62,14 +62,7 @@ public final class AuthenticationViewController<T: ServiceAuthenticationUI>: UIT
     }
```
gestrich, line 63: Return all credentials --whether authorized or not. This will allow us to show a field with ***** for the secret. Before this we would just show the username. This seemed helpful to remind users whether the service was setup with or without a secret, since it is an optional field. I'm thinking of the case specifically where a user decided to add or remove a secret field requirement on their Nightscout instance, then coming back into the app and not being sure why the Nightscout Api Client is failing.

MockKit/MockCGMManager.swift
```diff
@@ -299,7 +299,7 @@ public final class MockCGMManager: TestingCGMManager {
         return MockCGMManager.managerIdentifier
     }
```
gestrich, line 301: Most of our other services follow the convention <Name CGM>. Applying that pattern here to help with our sorted list of CGMs.

ps2: I think in this case simulator is the noun and CGM is the adjective, whereas in other cases CGM is the noun, so I think it should stay in this order.

ps2 (changes requested):

ps2:
To be more specific, I don't think "simulator" works as an adjective. I suppose you could change it to "*Simulated* CGM" and keep the adjective in front. But I still think CGM Simulator sounds more natural and describes what it is well.

ps2:
Thanks! I merged the ServiceAuthenticationUI changes in this here: https://github.com/LoopKit/LoopKit/commit/ece225720eabe8dda38ebcf69180872eb50308c2

gestrich:
@ps2 Sounds good. Those changes on the naming were per the suggestion of another dev in the community who suggested the naming across our list of CGMs was inconsistent affecting the sort ordering of the CGMs. It doesn't really matter either way to me and I understand why he suggested it and also why you prefer the existing name. Thanks.

PR #386, opened 2022-02-26
Lyumjev Support

damonbayer:


ngnrfsk:
Hey there Damon,

Thanks for adding this Lyumkev update. I see there's not onset time in this version of the Loop insulin absorption model. Is there a way to account for Lyumje's very rapid onset time?

Thanks,


iarla

damonbayer:
Can you elaborate? This is the exact same model as is being used for Fiasp, which seems to work well for people.

ngnrfsk:
I mean by comparison with 2.2.4, where there’s the freedom of another parameter to adjust, as follows in ExponentialInsulinModelPreset.swift

var effectDelay: TimeInterval {
        switch self {
        case .humalogNovologAdult:
            return .minutes(10)
        case .humalogNovologChild:
//            return .minutes(10)
            return .minutes(5) //lyumjev
        case .fiasp:
            return .minutes(10)
        }
    }

iarla

damonbayer:
The delay parameter is still there. It is 10 minutes for all insulin models: https://github.com/LoopKit/LoopKit/blob/dev/LoopKit/InsulinKit/ExponentialInsulinModel.swift#L27

Whether or not that is correct is up for debate, but I simply duplicated the Fiasp model and changed the name/color/description to be appropriate for Lyumjev. Seems to work well in practice.

ngnrfsk:
Thanks. I use the shorter delay for my use of lyumjev, as it can come on quite fast. Found when I was first setting the parameters from the kinetics it gave a slightly better result. Or possibly I imagined it. 
¯\_(ツ)_/¯ 

Off topic, is there a way to adjust the carb absorption rate?

Dr Iarla Kilbane-Dawe
+44 7739 804 892
***@***.***

damonbayer:
How would you like to adjust it? I find that dynamic carb absorption, multiple carb entries, and varying absorption times provide plenty of flexibility with respect to handling carbs.

ngnrfsk:
Fair question Damon. My thought was to have an absorption rate coefficient is a GI control, but practically I guess that’s fully covered by the absorption time coefficient. 

I’ll watch out for possible use cases and get back to you if I see anything different. 

Dr Iarla Kilbane-Dawe
+44 7739 804 892
***@***.***

PR #387, opened 2022-03-02
Afrezza Model

damonbayer:
Here are some first steps towards getting an Afrezza model into Loop.

Done:
- I fit an ExponentialInsulinModel to the [published data](https://www.accessdata.fda.gov/drugsatfda_docs/label/2014/022472lbl.pdf).
- I added an Afrezza type as an ExponentialInsulinModelPreset.

Known issues:
- Afrezza is shown as an option for pump insulin, but it should only show up as an external insulin.
- No tests for Afrezza doses.

Possible issue:
- I modified some aspects of the delay parameter in ExponentialInsulinModel, and I'm not sure if I did it correctly. It seems to work.


ps2:
Would you be up for creating a Loop (and maybe mdt/eros/dash) PR that removes afrezza from the list of `allowedInsulinTypes` passed to pumpmanagers for setup and settings? We may also need updates to the pumpmanagers themselves to respect the `allowedInsulinTypes` list.

damonbayer:
I tried my best at it for a few hours but couldn't get it to work. I really don't understand the code and am just pattern matching, which gets me a lot of errors that I don't understand. I would happily accept some hints, but it would be great to put this in the hands of someone who knows what they're doing.

damonbayer:
Thanks to @novalegra, Afrezza is now hidden from the pump insulins, but shows up in the external dose logging.

PR #388, opened 2022-03-14
include therapyTimeZone in schedules

marionbarker:
Add therapy time zones to schedules for tests.
With this change, running tests in Xcode with the Product->Test command, all tests pass.


ps2:
These are using the therapy timezone as the phone's simulated timezone, which is not how the schedules are actually used in practice. In practice the phone timezone is a real (not fixed offset) timezone that usually participates in DST. But this step should at least make the tests not fail.

PR #399, opened 2022-04-24
Expand Information in Therapy Setting and Expand Recommended Limit Ranges

marionbarker:
This has been closed in favor of single-topic PRs.
Adding these to the list as they are opened:
* PR #403 
* PR #404 

Purpose: 
1. Based on frequently asked questions, I've expanded some of the information provided in the Therapy Settings screens, primarily for Glucose Safety Limit, Correction Range and Pre-Meal Range.
2. When people who have had higher BG values begin to get better control (via MDI, manual pump, commercial closed loop or DIY systems), they should consult their medical team for advice on how rapidly they can safely reduce average BG.  

Discussion for item 1:
* I concentrated on getting information added
* If I broke formatting "rules" I am happy to be told how best to do it and then I will update my branch

Discussion for Item 2:
* Rapid lowering of A1C can exacerbate problems with retinopathy and neuropathy. 
* To accommodate new loopers, I increased the upper level ranges over which the glucose pickers remain black.
* I suggest we get review from an endocrinologist on the best wording and levels.

### DIFF
--- a/LoopKitUI/Views/Information Screens/CorrectionRangeOverrideInformationView.swift
+++ b/LoopKitUI/Views/Information Screens/CorrectionRangeOverrideInformationView.swift
@@ -51,7 +51,8 @@ public struct CorrectionRangeOverrideInformationView: View {
         case .preMeal:
             return VStack(alignment: .leading, spacing: 20) {
                 Text(String(format: LocalizedString("Your Pre-Meal Range should be the glucose value (or range of values) you want %1$@ to target by the time you take your first bite of your meal. This range will be in effect when you activate the Pre-Meal Preset button.", comment: "Information about pre-meal range format (1: app name)"), appName))
-                Text(LocalizedString("This will typically be", comment: "Information about pre-meal range relative to correction range")) + Text(LocalizedString(" lower ", comment: "Information about pre-meal range relative to correction range")).bold().italic() + Text(LocalizedString("than your Correction Range.", comment: "Information about pre-meal range relative to correction range"))
+                Text(LocalizedString("This will typically be", comment: "Information about pre-meal range relative to correction range")) + Text(LocalizedString(" lower ", comment: "Information about pre-meal range relative to correction range")).bold().italic() + Text(LocalizedString("than your Correction Range but cannot be lower than the Glucose Safety Limit.", comment: "Information about pre-meal range relative to correction range")) +
+                Text(LocalizedString("\n\nThe Pre-Meal duration is 1 hour or until user enters carbs, cancels Pre-Meal or enacts a different override.", comment: "Information about pre-meal duration"))
             }
             .fixedSize(horizontal: false, vertical: true) // prevent text from being cut off
         case .workout:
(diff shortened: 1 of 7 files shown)


LoopKitUI/Views/Information Screens/GlucoseTherapySettingInformationView.swift
```diff
@@ -96,7 +96,7 @@ fileprivate extension TherapySetting {
     }
 
     func lowHighText(lowerBoundString: String, upperBoundString: String) -> String {
-        return String(format: LocalizedString("It can be set as low as %1$@. It can be set as high as %2$@.",
+        return String(format: LocalizedString("Graphic above for illustration only; Code Limits:\n\tHigh Limit:\t%2$@\n\t Low Limit:\t%1$@",
```
novalegra, line 99: The formatting for this string causes it to spread a little weirdly across different screen sizes (I've included an example of what I see with an iPod Touch down below). I like the intention of clarifying the graphic for the user - I wonder if we could say something here along the lines of `"Graphic above for illustration only. This setting can be set as low as %1$@ and as high as %2$@."` in order to provide context about the graphic / setting while not running into formatting concerns

Captura de Pantalla 2022-04-25 a la(s) 8 51 16 a m

LoopKitUI/Views/Information Screens/CorrectionRangeOverrideInformationView.swift
```diff
@@ -51,7 +51,8 @@ public struct CorrectionRangeOverrideInformationView: View {
...
-                Text(LocalizedString("This will typically be", comment: "Information about pre-meal range relative to correction range")) + Text(LocalizedString(" lower ", comment: "Information about pre-meal range relative to correction range")).bold().italic() + Text(LocalizedString("than your Correction Range.", comment: "Information about pre-meal range relative to correction range"))
+                Text(LocalizedString("This will typically be", comment: "Information about pre-meal range relative to correction range")) + Text(LocalizedString(" lower ", comment: "Information about pre-meal range relative to correction range")).bold().italic() + Text(LocalizedString("than your Correction Range but cannot be lower than the Glucose Safety Limit.", comment: "Information about pre-meal range relative to correction range")) +
+                Text(LocalizedString("\n\nThe Pre-Meal duration is 1 hour or until user enters carbs, cancels Pre-Meal or enacts a different override.", comment: "Information about pre-meal duration"))
```
novalegra, line 55: [nit] Instead of adding newlines at the beginning of the "The Pre-Meal duration" string, I wonder if the `+` between the two `Text`s  could be eliminated to achieve the same spacing while being a little clearer in the code. 

I know this causes a SwiftUI opaque return type error - do others have a sense of why that might be happening here? It doesn't fall into the 2 more common cases (forgetting to return the type or views not being in a container)

novalegra (commented):
I've left a few comments about string formatting but this PR overall looks in good shape to me.  I don't have the medical background to comment on the guardrail changes, so someone with that should review those before this is merged in.

ps2:
Thanks Marion. The current guardrails were picked based on review of the Loop observational study data, and review by endocrinologists. The wording has also been reviewed. I'm not saying this to say they can't change, but we need clear evidence and good reasoning behind each one. It would also help to split this up into smaller PRs; many of the items here are related, I know, but most of them could be taken independently, and that would help with getting them through the process. Some of them are easy to take, and some require more discussion/evidence.

marionbarker:
Will do.

gigich:
Hi
Normaly I change the Glucose Safety Limit to 54 (3.0 mmol/L), because otherwise if we enter carbs on low bg, it will not deliver insulin before meals.

marionbarker:
I am closing this PR because I will be opening different pull requests that focus on single changes in this topic area.

PR #400, opened 2022-04-25
add a reset button to Pre-Meal screen

marionbarker:
This is in response to [Issue 1654](https://github.com/LoopKit/Loop/issues/1654) (listed under LoopKit/Loop).

This code adds a Reset button to upper right of the Pre-Meal Settings screen, shown in graphic below.
Action of the button has been tested.

PreMeal_ResetButton

### DIFF
--- a/LoopKitUI/Views/Settings Editors/CorrectionRangeOverridesEditor.swift
+++ b/LoopKitUI/Views/Settings Editors/CorrectionRangeOverridesEditor.swift
@@ -86,6 +86,11 @@ public struct CorrectionRangeOverridesEditor: View {
         content
             .navigationBarBackButtonHidden(shouldAddCancelButton)
             .navigationBarItems(leading: leadingNavigationBarItem)
+            .navigationBarItems(leading: EmptyView(), trailing: resetButton)
+    }
+
+    private var resetButton: some View {
+        Button(action: { value.ranges = [:] } ) { Text(LocalizedString("Reset", comment: "Restore settings to nil")) }
     }
     
     private var cancelButton: some View {
--- a/LoopKitUI/Views/Settings Editors/CorrectionRangeOverridesEditor.swift
+++ b/LoopKitUI/Views/Settings Editors/CorrectionRangeOverridesEditor.swift
@@ -85,15 +85,15 @@ public struct CorrectionRangeOverridesEditor: View {
     private var contentWithCancel: some View {
         content
             .navigationBarBackButtonHidden(shouldAddCancelButton)
-            .navigationBarItems(leading: leadingNavigationBarItem, trailing: unsetButton)
+            .navigationBarItems(leading: leadingNavigationBarItem, trailing: deleteButton)
     }
 
-    private var unsetButton: some View {
+    private var deleteButton: some View {
         Button( action: {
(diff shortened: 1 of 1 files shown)


LoopKitUI/Views/Settings Editors/CorrectionRangeOverridesEditor.swift
```diff
@@ -85,7 +85,11 @@ public struct CorrectionRangeOverridesEditor: View {
     private var contentWithCancel: some View {
         content
             .navigationBarBackButtonHidden(shouldAddCancelButton)
-            .navigationBarItems(leading: leadingNavigationBarItem)
+            .navigationBarItems(leading: leadingNavigationBarItem, trailing: resetButton)
+    }
+
+    private var resetButton: some View {
+        Button(action: { value.ranges = [:] } ) { Text(LocalizedString("Reset", comment: "Restore settings to nil")) }
```
ps2: "Reset" implies reverting to some previous version, which is a little unclear, I think. Probably something like "delete", "remove", or "unset" would be better.

ps2 (changes requested):
Thanks! Just a comment on the wording.

marionbarker:
I modified the button name to `Unset`.

I think the button action needs one more thing and I don't know how to do it.

If the user begins to adjust settings, so the selection dials are active and then the user taps `Unset`, the values return to nil ([:]) but the display of selection dial remains active as shown in the graphic below.

I believe after tapping Unset, the display of the selection dials should be turned off.

premeal_unset_when_dials_are_displayed

marionbarker:
Thanks again to Anna for the hint.  I believe this PR is ready for another review.

LoopKitUI/Views/Settings Editors/CorrectionRangeOverridesEditor.swift
```diff
@@ -85,7 +85,15 @@ public struct CorrectionRangeOverridesEditor: View {
...
-            .navigationBarItems(leading: leadingNavigationBarItem)
+            .navigationBarItems(leading: leadingNavigationBarItem, trailing: unsetButton)
+    }
+
+    private var unsetButton: some View {
+        Button( action: {
+                value.ranges = [:];
+                presetBeingEdited = nil
+        } )
+        { Text(LocalizedString("Unset", comment: "Restore settings to nil")) }
```
ps2: Heh, that was my least favorite of the three choices.  I just included it for completeness, but now I'm thinking that having the word "Unset" as standalone text (albeit blue colored) might read more like an adjective, rather than a verb. I won't block it on that if you think this is the right choice for some reason, but I do think "Delete" or "Clear" better communicates that the text is an action to take.  Unset is also fairly rare as a verb, I think.

marionbarker: Changed button name to Delete.

BTW - there is comment text associated with that button.  When is that used?  Is it part of accessibility access?
I modified the comment as well.

ps2: It's used to give more information about the context of the string to translators.

ps2:
Thanks!

PR #401, opened 2022-04-25
Report Total Daily Scheduled Basal

marionbarker:
Several people have requested that the daily scheduled basal rate be reported (as it was in prior versions of Loop).
This has been discussed in zulipchat under [#development: Total daily basal](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Total.20daily.20basal/near/278302709)
There is a (very old) open issue requesting this as well [LoopKit/Loop Issue 841](https://github.com/LoopKit/Loop/issues/841)

Use Case:
1. After modifying the basal schedule, seeing the total assists user in verifying no improper entry was made
2. Many HCP ask for total daily basal rates - with closed-loop systems, they probably need the Scheduled Basal Total

### DIFF
--- a/LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
+++ b/LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
@@ -264,9 +264,11 @@ extension TherapySettingsView {
 
     private var basalRatesSection: Card {
         card(for: .basalRate) {
-            if let items = viewModel.therapySettings.basalRateSchedule?.items,
-               let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates,
-               let total = viewModel.therapySettings.basalRateSchedule?.total()  {
+            if let schedule = viewModel.therapySettings.basalRateSchedule,
+               let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates
+            {
+                let items = schedule.items
+                let total = schedule.total()
                 SectionDivider()
                 ForEach(items.indices, id: \.self) { index in
                     if index > 0 {
--- a/LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
+++ b/LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
@@ -267,7 +267,7 @@ extension TherapySettingsView {
             if let items = viewModel.therapySettings.basalRateSchedule?.items,
                let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates,
                let total = viewModel.therapySettings.basalRateSchedule?.total()  {
-                Text(LocalizedString("\(round(total * 100) / 100.0) U/day", comment: "Schedule Basal Daily Total"))
+                Text(LocalizedString("\(round(total * 100) / 100.0) U/day", comment: "Scheduled Basal Daily Total"))
                 SectionDivider()
                 ForEach(items.indices, id: \.self) { index in
                     if index > 0 {
--- a/LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
+++ b/LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
@@ -267,7 +267,6 @@ extension TherapySettingsView {
             if let items = viewModel.therapySettings.basalRateSchedule?.items,
                let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates,
                let total = viewModel.therapySettings.basalRateSchedule?.total()  {
-                Text(LocalizedString("\(round(total * 100) / 100.0) U/day", comment: "Scheduled Basal Daily Total"))
                 SectionDivider()
                 ForEach(items.indices, id: \.self) { index in
                     if index > 0 {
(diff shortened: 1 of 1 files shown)


LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
```diff
@@ -264,7 +264,10 @@ extension TherapySettingsView {
 
     private var basalRatesSection: Card {
         card(for: .basalRate) {
-            if let items = viewModel.therapySettings.basalRateSchedule?.items, let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates {
+            if let items = viewModel.therapySettings.basalRateSchedule?.items,
+               let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates,
+               let total = viewModel.therapySettings.basalRateSchedule?.total()  {
+                Text(LocalizedString("\(round(total * 100) / 100.0) U/day", comment: "Scheduled Basal Daily Total"))
```
ps2: This needs to be a format string, for localization. See other examples where String(format...) is used

LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
```diff
@@ -264,7 +264,10 @@ extension TherapySettingsView {
 
     private var basalRatesSection: Card {
         card(for: .basalRate) {
-            if let items = viewModel.therapySettings.basalRateSchedule?.items, let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates {
+            if let items = viewModel.therapySettings.basalRateSchedule?.items,
+               let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates,
+               let total = viewModel.therapySettings.basalRateSchedule?.total()  {
```
ps2: This reads somewhat like you do not want to display any part of the card if you can't display the total.  I.e. it suggests to the reader that total() might return nil, but really it's just the schedule that might be nil.
```
if let schedule = viewModel.therapySettings.basalRateSchedule,  let supportedBasalRates = viewModel.pumpSupportedIncrements()?.basalRates {
    let items = schedule.items
    let total = schedule.total()
```

ps2 (changes requested):
I haven't checked this out to see how it displays yet. It does need a format string change for localization.  Maybe attach a screenshot if you could?

LoopKitUI/Views/Settings Editors/TherapySettingsView.swift
ps2: "After modifying the basal schedule, seeing the total assists user in verifying no improper entry was made".

If you're trying to address this use case, maybe showing it on the editing screen would be better.  Also, the other use case is rarer occurance (once every 1-6mo), so might be better to have it in the context of the editing screen also, rather than adding cognitive load every time the user goes to settings.

marionbarker: I wanted to also display the total with each modification to a given entry but I don't know how.
I do know how to put the total at the top of the list or the bottom of the list.

Use case for showing the total on the Settings Screen:
* I believe it simplifies ability to get an overview of scheduled basal rate.
* Do you want more use-case explanations for this here or in the zulipchat conversation?

(and IMO, it does not increase cognitive load to have the daily total present)

I'll add some screen shots in a future comment - and also test it with a fresh install.

And I'll figure out how to do the localizations correctly.

marionbarker:
Graphics with display on top, bottom and for a fresh install.

The final graphic shows a filled out basal schedule during on-boarding (simulator) with several entries, reinforcing the need to figure out how to display it on this screen.

TherapySettings_BasalSchTotal_Top

TherapySettings_BasalSchTotal_Bottom

TherapySettings_Basal_FreshInstall

TherapySettings_Basal_FreshInstall_OnboardingSeveralEntries

ps2:
Thanks for the screenshots. It should probably match the structure of the other rows.
So "Total" on the left, primary color, then spacing, then the total value in primary color, and then "U/day" in secondary color. 

And the bottom is the right place for it, as a summation.

marionbarker:
I have a WIP update.  With this different way of creating the text string, the `\(round(total * 100) / 100.0)` used in commit [2762e10](https://github.com/LoopKit/LoopKit/pull/401/commits/2762e10dce4ee7638ee95ffcb0f3c0052820882e) no longer provides the correct formatting.

I decided to post the wip commit and ask for help. 
Note - I exceeded one request and made Total bold - happy to remove that if you prefer. I was playing around with formatting and left it in for now.

If I figure out how to add this into to the display when editing basal rate, I will make a separate PR.

This is the WIP display with commit [3d4aaa7](https://github.com/LoopKit/LoopKit/pull/401/commits/3d4aaa7ed04df028a2db818857505e376ee18011)

Basal_Daily_Total_WIP

marionbarker:
Realized I had not incorporated this suggested change (https://github.com/LoopKit/LoopKit/pull/401/files#r857840252).
This was added with commit 7a86a40.  No change to display.

marionbarker:
Realized I had not incorporated this suggested change (https://github.com/LoopKit/LoopKit/pull/401/files#r857840252).
This was added with commit 7a86a40.  No change to display.

marionbarker:
I'm reporting total daily basal to 2 decimal places now. Let me know if you want any other changes.

This still has the word total as bold.

Basal_Daily_Total_DecimalLimit

ps2:
I think this looks good, and the bold is fine. Only tiny tweak is maybe that we don't need two digits of precision in a lot of cases (and in the one shown).

I believe the previous code used something like this:

```
let formatter = NumberFormatter()

formatter.numberStyle = .decimal
formatter.minimumFractionDigits = 0
formatter.maximumFractionDigits = 2

formatter.string(from: 25.50)
```
Which would result in "25.5", instead of "25.50".

marionbarker:
If you feel strongly about this change, then I'll make it. 
(In my mind, the number of decimal places is an indication of the precision with which you know a quantity, so 2 decimal places (or 3 for some pumps) is appropriate for every line.)

My preference is that this get merged as an interim fix to something many folks have requested and that I feel strongly needs to be available.
When I figure out how to add this to the screen where entries are being edited, we can adjust the formatting in both places.

ps2:
Yeah, I don't think removing the unnecessary precision is a blocker, but I do think it'd be good to switch back to the original formatting when we can.

ps2:
Thanks!

PR #403, opened 2022-04-27
Increase upper range of recommendedBounds

marionbarker:
Increase upper range of recommendedBounds for suspendThreshold and correctionRange.

This provides support to new Loopers who may wish to start at higher levels than is typical.

The overall ranges (absoluteBounds) are not modified.

### DIFF
--- a/Extensions/Guardrail+Settings.swift
+++ b/Extensions/Guardrail+Settings.swift
@@ -9,7 +9,11 @@
 import HealthKit
 
 public extension Guardrail where Value == HKQuantity {
-    static let suspendThreshold = Guardrail(absoluteBounds: 67...110, recommendedBounds: 74...80, unit: .milligramsPerDeciliter, startingSuggestion: 80)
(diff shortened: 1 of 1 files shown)


ps2:
This needs to be done in a code-configurable way, as Tidepool can't take these changes easily.

marionbarker:
Happy to make changes but I’m not sure how you want it changed. Do you mean create (or use an existing) config file to hold the values?

ps2:
The ideal solution for this is 1) have these values not specified in LoopKit, but allow Loop to control them. 2) Be able to override the default Loop values at build time, preferably via something like xcconfig (LoopConfigOverride), so custom guardrails could be specified at a higher level. But xcconfig isn't great for specifying more complex types, so I'm not quite sure what the best solution here is. We definitely need build time config.

ps2:
Closing this PR. Not opposed to allowing a higher range, but there is a lot more to supporting that than what's submitted in this PR. Graphic in therapy settings also affected.

PR #404, opened 2022-04-27
Enhance Therapy Information Screens

marionbarker:
There will be a series of commits using this branch for this Pull Request.

Each commit will modify a single information screen. (Well that was the idea, but of course that got changed as I worked through them).
A graphic will be added to the conversation for each screen as updated.

* modified BasalRatesInformationView.swift

Minor wording change to 3rd phrase.  
Addition of 4th and 5th phrase.

BasalRatesInformationView_Update

### DIFF
--- a/LoopKit/TherapySetting.swift
+++ b/LoopKit/TherapySetting.swift
@@ -17,6 +17,8 @@ public enum TherapySetting: String {
     case insulinModel
     case carbRatio
     case insulinSensitivity
+    case addNewEntry
+    case editDeleteEntry
     case none
 }
 
@@ -43,7 +45,7 @@ public extension TherapySetting {
             return LocalizedString("Carb Ratios", comment: "Title text for carb ratios")
         case .insulinSensitivity:
             return LocalizedString("Insulin Sensitivities", comment: "Title text for insulin sensitivity")
-        case .none:
+        case .none, .addNewEntry, .editDeleteEntry :
             return ""
         }
     }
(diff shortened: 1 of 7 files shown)


marionbarker:
Update for the CarbRatioInformationView screen.
Minor edit of 2nd phrase.
Addition of 3rd phrase.

CarbRatioInformationView_Update

marionbarker:
I'm going to pause after this update and request feedback.  All the graphics have been update to the latest commit.

There were common phrases used more than once.  I added them as new cases in `LoopKit/TherapySetting.swift` which required modification of all code that used that enum list.

I believe the repeated strings should be common - but this may not be the correct place for them to land.
I used the two strings again in the updated Correction Range Information screen shown here. The other modification was to use `insulin delivery` instead of `basal insulin` and to make Consult your health care provider disclaimer bold.


CorrectionRangeInformationView_Update

LoopKit/TherapySetting.swift
```diff
@@ -17,6 +17,8 @@ public enum TherapySetting: String {
     case insulinModel
     case carbRatio
     case insulinSensitivity
+    case addNewEntry
```
ps2, line 20: This enum represents different Types of settings, not different actions on those settings. It's not a general place to store translation strings, so these two new enum values are not appropriate here.

LoopKitUI/Views/Information Screens/BasalRatesInformationView.swift
```diff
@@ -34,8 +34,10 @@ public struct BasalRatesInformationView: View {
         VStack(alignment: .leading, spacing: 25) {
             Text(LocalizedString("Your Basal Rate of insulin is the number of units per hour that you want to use to cover your background insulin needs.", comment: "Information about basal rates"))
             Text(String(format: LocalizedString("%1$@ supports 1 to 48 rates per day.", comment: "Information about max number of basal rates"), appName))
-            Text(LocalizedString("The schedule starts at midnight and cannot contain a rate of 0 U/hr.", comment: "Information about basal rate scheduling"))
-        }
+            Text(LocalizedString("The schedule starts at midnight and cannot contain a single rate of 0 U/hr.", comment: "Information about basal rate scheduling"))
```
ps2, line 37: This is pump specific. Until we can adapt and provide pump specific advice, the lowest common denominator functionality should probably be described here.

ps2 (changes requested):

marionbarker:
I'll put these changes into LoopDocs instead of trying to add them to the app information screens.

Issue #398, opened 2022-04-17
Loop crashes due to assert


tazitoo:
Running Loop v2.3.0 (recently built with Xcode 13.3.1) - it will crash.  Plugging iPhone into computer, rebuilding, and running through Xcode, shows the cause to be in Loopkit > InsulinKit > InsulinMath > reconciled() line 376.

If I comment out that line and rebuild - Loop runs fine.  When it crashes, and the phone is attached, you can see in the Xcode debugger lower left panel that `last.isMutable` is `True`.

Apologies, since I am not fluent in Xcode, Loop or Swift - just reporting what I am experiencing.

kuryk322:
I'm seeing the same thing, my app worked for over 24 hours before it hit this crash.  I reported it here - https://github.com/LoopKit/Loop/issues/1674#issue-1206518089

PieterGit:
Issue has been fixed in Dev. This issue may be closed. See https://github.com/LoopKit/LoopKit/pull/406#issuecomment-1146788184 for details.

PR #406, opened 2022-05-03
workaround for Loop crashing on dev (2.3.0) due to assert(!suspend.isMutable)

PieterGit:
This is a workaround for Loop crashing on dev branch. 
This has been reported by @tazitoo, @kuryk322, @elnjensen and myself.

See discussion in:
* https://github.com/LoopKit/Loop/issues/1674
* https://github.com/LoopKit/LoopKit/issues/398

Note that is one more case on line 389, which I did not fix, because it didn't crash on that line for me.

```
  case .resume:
                if let suspend = lastSuspend {
                    assert(!suspend.isMutable)
```

### DIFF
--- a/LoopKit/InsulinKit/InsulinMath.swift
+++ b/LoopKit/InsulinKit/InsulinMath.swift
@@ -373,8 +373,6 @@ extension Collection where Element == DoseEntry {
                 reconciled.append(dose)
             case .basal, .tempBasal:
                 if lastSuspend == nil, let last = lastBasal {
-                    // assert(!last.isMutable)
-
                     let endDate = Swift.min(last.endDate, dose.startDate)
 
                     // Ignore 0-duration doses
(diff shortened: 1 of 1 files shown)


PieterGit:
@ps2 Could you please review this fix? And tell us if it's safe/OK to remove this `assert`

PieterGit:
Thanks to @marionbarker who pointed me that this was reported by @elnjensen, see:
https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Assertion.20failure.20with.20Medtronic.20pump

I changed the PR to remove these asserts (and the extra whitespace). I hope @ps2 can confirm this is safe to do so.
This issues causes Loop crashes on MDT 522 and MDT 754 pumps with current Loop dev.

elnjensen:
I guess the question here is whether @ps2 thinks those checks are useful for Omnipod troubleshooting - they were put in as he was working on an issue on the Dash side of things.   If they are helpful for that, we’ll need some pump-specific code.  But if that’s the case, they probably shouldn’t live in InsulinMath, but instead somewhere in the pump manager code, I’d guess.

PieterGit:
Has been fixed by @ps2 with https://github.com/LoopKit/LoopKit/commit/af264fde95730142349b272e4fe3cc40d185a8bd
PR can be closed. Dev works again with MDT 522 and MDT 754 pumps. Thanks!

PR #414, opened 2022-06-25
Allow three decimal places in basal rate picker on supported pumps

elnjensen:
This small change allows up to three decimal places in the basal rate picker, for supported pumps, such as the Medtronic 523 which allows 0.025 U/hr basal rate increments.  Addresses issue LoopKit/Loop#1731.


ps2:
I think we should find a way to do this without making the picker for pod rates show three digits. `FractionalQuantityPicker` could scan the `selectableValues` array and make a determination based on the precision of the values it sees there.

marionbarker:
I have this modification on my real phone with Omnipod DASH: 2 decimal places for basal rates.

Test phone with that one-line change, look at basal rate picker:
* Omnipod (eros): 2 decimal places
* MDT 515:  2 decimal places
* MDT 523: Have to go below 1 U to see 3 decimals, but then I see it.
* MDT 515 (again): Had trouble reconnecting - kept saying max limits exceeded. But I know they are not (set them higher anyway). Also got decoding error. But consistently getting max limits exceeded.
Giving up on this pump - it's not cooperating.

marionbarker:
I had left a basal rate of 0.675 U/hr in the basal schedule when testing the MDT 523.
Added a simulator pump, changed basal rate to multiple of 0.05 and was able to connect to MDT 515 again.
I'll write up another issue for this.

elnjensen:
Coming back to this, as I somehow missed Pete’s comment the first time around.  As Marion notes, the way the code is written (not my change, but what was there before) it only shows as many digits as are needed in order to capture the enumerated basal rates.  

But previously, two is the max it would ever show. With the change, that _max_ is three, but it doesn’t show three digits unless the last one is significant, e.g. for x23 pumps.  So I think is the right way to go.  I don’t have an x23 to test, but Marion’s test sounds ok - except I guess that we don’t see 3 places unless below 1.  That’s not quite right.

ps2:
Thanks @elnjensen! Sorry for the delay on this. Just needed to find some time to test it. Looks good.

PR #428, opened 2022-11-08
Add missed meal notifications to Loop

novalegra:
This PR and the [associated Loop PR](https://github.com/LoopKit/Loop/pull/1825) add missed meal notifications to Loop. 

## Motivation
It's easy to forget to log a meal, and the Loop algorithm works best when it has information about all the carbs and insulin that could affect blood glucose levels. This PR uses the insulin dosing information in Loop to automatically detect when unexpected rise in BG is so large that the only likely explanation is a missed meal, and notify the user so they can log the meal and administer insulin if needed.

## Changes Made
Add helper utilities to support the `MealDetectionManager` in Loop.


LoopKit/Extensions/NSUserActivity+CarbKit.swift
```diff
@@ -13,17 +13,19 @@ extension NSUserActivity {
     public static let newCarbEntryActivityType = "NewCarbEntry"
 
     public static let newCarbEntryUserInfoKey = "NewCarbEntry"
+    public static let carbEntryIsUAMUserInfoKey = "CarbEntryIsUAM"
```
ps2: Hmm, this class looks like it should probably move to Loop, now that the carb entry interface lives there. Not something you need to do for this ticket; not really related; I just noticed.

LoopKit/CarbKit/CarbStore.swift
```diff
@@ -1432,6 +1446,149 @@ extension CarbStore {
     }
 }
 
+// MARK: Missed / Unannounced Meal Detection
+extension CarbStore {
+    public func hasUnannouncedMeal(insulinCounteractionEffects: [GlucoseEffectVelocity], completion: @escaping (UnannouncedMealStatus) -> Void) {
```
ps2: I wonder if this could be moved out of CarbStore. Seems like a meal detector would consume data from CarbStore, but otherwise be fairly independent of it.

LoopKit/LoopMath.swift
```diff
@@ -46,6 +46,35 @@ public enum LoopMath {
...
+    /**
+     Calculates a range of time in `delta`-value intervals
+
+     - parameter start:     The range start date
+     - parameter end:       The range end date
+     - parameter delta:     The time differential for items in the returned range
+
+     - returns: An array of dates
+     */
+    static func dateRange(
```
ps2: maybe call this a "simulationDateRange"

LoopKit/Extensions/NSUserActivity+CarbKit.swift
novalegra: I can mark it as a FIXME

LoopKit/CarbKit/CarbStore.swift
novalegra: Done! I've moved it into Loop as its own manager (which owns both the meal detection algorithm & notifications)

PR #429, opened 2022-11-17
Fix Delete for MockCGM, was being retained

marionbarker:
Problem: Delete of MockCGM did not work
* Old simulator kept running in addition to any new CGM added to system

Solution: change `let` to `weak var`


ps2:
Thanks, but I don't think this is the right fix. `MockCGMManagerSettingsViewController` should be able to have a strong reference to `MockCGMManager`, as the CGMManager should not go away while the VC is on screen.  What should happen is when the VC goes away, any references to the CGMManager should be released. But that's not happening... I'll take a look.

marionbarker:
Use the fix posted in PR #430

PR #431, opened 2022-11-17
Add ability to estimate delivery duration of bolus

novalegra:
I added a function to the `PumpManager` protocol to support estimating the duration to deliver a bolus. I chose not to add it to `DoseProgress` / `DoseProgressReporter` since I wanted to provide a lightweight function to compute the duration and `DoseProgressReporter` has a lot of other functionality (timers, observers, etc) that isn't needed for this specific usecase. We also already provide lightweight functions on `PumpManager` for computing things like the rounding to the specific bolus volume the pump supports.

See https://github.com/LoopKit/Loop/pull/1830, https://github.com/LoopKit/OmniBLE/pull/68, and https://github.com/ps2/rileylink_ios/pull/751 for other associated changes.

### DIFF
--- a/LoopKit/DeviceManager/PumpManager.swift
+++ b/LoopKit/DeviceManager/PumpManager.swift
@@ -169,7 +169,7 @@ public protocol PumpManager: DeviceManager {
     func createBolusProgressReporter(reportingOn dispatchQueue: DispatchQueue) -> DoseProgressReporter?
 
     /// Returns the estimated time for the bolus amount to be delivered
-    func estimatedDuration(toDeliver bolusUnits: Double) -> TimeInterval
+    func estimatedDuration(toBolus units: Double) -> TimeInterval
     
     /// Send a bolus command and handle the result
     ///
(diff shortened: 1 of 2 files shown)


LoopKit/DeviceManager/PumpManager.swift
```diff
@@ -168,6 +168,9 @@ public protocol PumpManager: DeviceManager {
     /// Returns a dose estimator for the current bolus, if one is in progress
     func createBolusProgressReporter(reportingOn dispatchQueue: DispatchQueue) -> DoseProgressReporter?
 
+    /// Returns the estimated time for the bolus amount to be delivered
+    func estimatedDuration(toDeliver bolusUnits: Double) -> TimeInterval
```
ps2: Should "bolus" be part of the calling signature somehow?  Might be problematic if we want to define a function to estimate other types of delivery.

ps2 (approved):
LGTM, with one suggestion.

LoopKit/DeviceManager/PumpManager.swift
novalegra: Agreed! Perhaps `func estimatedDuration(toBolus units: Double) -> TimeInterval`?

PR #436, opened 2022-12-06
Fix Loop-dev crash from Correction Range accepted by Onboarding

SwiftlyNoah:
Addressing the [issue](https://github.com/LoopKit/Loop/issues/1853) posted by Marion. Created a new extension with functions to ensure that the values presented in the UI are at minimum equal to the minValue (the maximum of 87 and the suspend threshold). This will prevent the bounds crash noted in Marions issue report.

### DIFF
--- a/LoopKit/GlucoseRangeSchedule+SafeBounds.swift
+++ b/LoopKit/GlucoseRangeSchedule+SafeBounds.swift
@@ -10,18 +10,14 @@ import Foundation
 import HealthKit
 
 extension GlucoseRangeSchedule {
-    public func safeSchedule(with suspendThreshold: GlucoseThreshold?) -> GlucoseRangeSchedule? {
-        var glucoseScheduleMinimunValue: Double!
-        if let threshold = suspendThreshold {
-            glucoseScheduleMinimunValue = Guardrail.minCorrectionRangeValue(suspendThreshold: threshold).doubleValue(for: threshold.unit)
-        }
-        else {
-            glucoseScheduleMinimunValue = Guardrail.correctionRange.absoluteBounds.lowerBound.doubleValue(for: .milligramsPerDeciliter)
-        }
-        return rangeScheduleWithMin(min: glucoseScheduleMinimunValue)
-    }
-
-    private func rangeScheduleWithMin(min: Double) -> GlucoseRangeSchedule? {
+    public func safeSchedule(with suspendThreshold: Double?, unit: HKUnit) -> GlucoseRangeSchedule? {
+        var min: Double!
+        min = [
+            suspendThreshold,
+            Guardrail.correctionRange.absoluteBounds.lowerBound.doubleValue(for: unit)
+        ]
+            .compactMap({ $0 })
+            .max()
         let filteredItems = rangeSchedule.valueSchedule.items.map { scheduleValue in
             let newScheduleValue = DoubleRange(minValue: max(min, scheduleValue.value.minValue), maxValue: max(min, scheduleValue.value.maxValue))
             return RepeatingScheduleValue(startTime: scheduleValue.startTime, value: newScheduleValue)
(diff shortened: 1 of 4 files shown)


LoopKit/GlucoseRangeSchedule+SafeBounds.swift
```diff
@@ -0,0 +1,34 @@
...
+        }
+        else {
+            glucoseScheduleMinimunValue = Guardrail.correctionRange.absoluteBounds.lowerBound.doubleValue(for: .milligramsPerDeciliter)
+        }
+        return rangeScheduleWithMin(min: glucoseScheduleMinimunValue)
+    }
+
+    private func rangeScheduleWithMin(min: Double) -> GlucoseRangeSchedule? {
+        let filteredItems = rangeSchedule.valueSchedule.items.map { scheduleValue in
+            let newScheduleValue = DoubleRange(minValue: max(min, scheduleValue.value.minValue), maxValue: max(min, scheduleValue.value.maxValue))
```
ps2: The units of the value passed in here, `min`, are mg/dL.  The scheduleValue.value items are going to be in units specified by `unit`.

ps2 (changes requested):
Thanks Noah! I think there is a unit conversion issue here that needs to be addressed.

LoopKit/GlucoseRangeSchedule+SafeBounds.swift
SwiftlyNoah: Thanks for catching that, I just fixed it. Also, the safeSchedule function will now only be called in onboarding, as its not necessary to check if a user is just modifying their settings.

LoopKitUI/ViewModels/CorrectionRangeScheduleEditorViewModel.swift
```diff
@@ -20,10 +20,17 @@ struct CorrectionRangeScheduleEditorViewModel {
...
-         didSave: (() -> Void)? = nil)
-    {
-        self.glucoseTargetRangeSchedule = therapySettingsViewModel.glucoseTargetRangeSchedule
+    init(
+        mode: SettingsPresentationMode,
+        therapySettingsViewModel: TherapySettingsViewModel,
+        didSave: (() -> Void)? = nil
+    ) {
+        if mode == .acceptanceFlow, let unsafeSchedule = therapySettingsViewModel.glucoseTargetRangeSchedule {
+            self.glucoseTargetRangeSchedule = unsafeSchedule.safeSchedule(with: therapySettingsViewModel.suspendThreshold?.value, unit: unsafeSchedule.unit)
```
ps2: `therapySettingsViewModel.suspendThreshold` has its own, separate, units from `unsafeSchedule.unit` here. I'd recommend making `safeSchedule` take an HKQuantity, and you can then pass in `therapySettingsViewModel.suspendThreshold.quantity` to `safeSchedule`.  `safeSchedule`, since it is a method on `GlucoseRangeSchedule`, has access to `unit` locally, and doesn't need to have units passed in.

marionbarker:
Test for guardrails; incoming values < min allowed was OK.
Broke with test: incoming values > max allowed when onboarding.

* Delete Loop-dev from phone
* Modify FreeAPS to allow higher correction range
* Build FreeAPS
   * set suspend at 120 and correction range at 200 to 200
Build Loop-dev with PR 436 patch up through commit 346a1f3
* onboarding: 
   * safety limit could not be confirmed until I modified from 120 to 110
   * correction range allowed me to confirm.

Subsequent crash when attempting to edit correction range.

Screenshot 2022-12-06 at 9 54 35 AM

SwiftlyNoah:
You might be the only one to think of that Marion :)

Fixed!

marionbarker:
Applied updated patch to LoopWorkspace commit 9fd16c2.
Started with FreeAPS with suspend at 60 mg/dL, correction at 70 to 200 mg/dL.
* Glucose Suspend started at 67 mg/dL (min guardrail)
* Correction range started at 87 to 180 mg/dL (min and max guardrails)
* Success but I have a UX comment that I'll put into zulipchat
* I like the UX for this change better than the existing ones for Glucose Safety and Delivery Limits.

LoopKitUI/ViewModels/CorrectionRangeScheduleEditorViewModel.swift
SwiftlyNoah: I just implemented those changes!

PR #437, opened 2022-12-10
Improve Remote failure error messaging

gestrich:


LoopKit/Service/RemoteDataService.swift
```diff
@@ -118,5 +118,16 @@ public extension RemoteDataService {
     var pumpEventDataLimit: Int? { return nil }
     var settingsDataLimit: Int? { return nil }
```
gestrich, line 120: Replace the signature with a Result type so I can capture the Error. I'm partial towards throwing APIs in Swift these days but I see Result types more common throughout Loop.

LoopKit/Notification/LoopNotificationCategory.swift
```diff
@@ -20,6 +20,7 @@ public enum LoopNotificationCategory: String {
     case alert
     case remoteBolus
     case remoteBolusFailure
+    case remoteCommandExpired
```
ps2, line 23: This and the type above should probably go away when error reporting is being done correctly in the caregiver app.

LoopKit/Service/RemoteDataService.swift
ps2: When the error path is correctly feeding errors to the caregiver app, the flow of errors here will be obviated as well.

ps2 (approved):
As much as I don't like adding more code that will eventually have to be rewound, I can understand this making debugging easier for now. I would hope some progress can be made towards the right path soon, as the more we invest in this workaround, the more work the detangling later is going to be.

PR #449, opened 2023-02-03
Store the history without the conflicting overrides

bjorkert:
Overlapping overrides was filtered out, but the cleaned list was never stored/persisted before exiting. 

Next time loop starts, it will crash with the same error.


bjorkert:
Relates to https://github.com/LoopKit/Loop/issues/1763

ps2:
I'm a little torn between leaving this out so we have a better chance of replicating it, and fixing the underlying bug that causes it. But I'm going to merge it, as it seems rare, and could help prevent someone from getting in a bad situation if they do encounter it.

ksilive:
thanks, solved my loop crashing indefinitely

PR #454, opened 2023-02-22
Reorder Therapy Settings cards to match onboarding order

marionbarker:
I noticed that the order in which Therapy Settings are presented during Onboarding is different from the order of cards on the Therapy Settings screen after onboarding.

I moved the Carb Ratio card to be just before the Insulin Sensitivity Factor screen to match the Onboarding order, which is:
* Glucose Safety Limit
* Correction Range
* PreMeal Range
* Basal Rate
* Delivery Limits
* Carb Ratio
* Insulin Sensitivity Factor


ps2:
If we change anything, it should be the onboarding order.  The order in therapy settings matches Tidepool, and Tidepool's onboarding matches therapy settings.

PR #455, opened 2023-02-25
Remote PR Set #2: Introduce RemoteCommands

gestrich:
This PR is part of the below set that should be merged together. This one will use "RemoteCommands", a protocol for interacting with the remote services that is the source of the RemoteCommand. This will lay the groundwork for remote 2.0 commands.

* Loop: https://github.com/LoopKit/Loop/pull/1934
* LoopKit: https://github.com/LoopKit/LoopKit/pull/455
* NightscoutService: https://github.com/ps2/NightscoutService/pull/33
* TidepoolService: https://github.com/LoopKit/TidepoolService/pull/4

### DIFF
--- a/LoopKit/Service/Remote/RemoteCommand.swift
+++ b/LoopKit/Service/Remote/RemoteCommand.swift
@@ -1,0 +1,13 @@
+//
+//  RemoteCommand.swift
+//  LoopKit
+//
+//  Created by Bill Gestrich on 2/25/23.
+//  Copyright © 2023 LoopKit Authors. All rights reserved.
+//
+
+public protocol RemoteCommand {
+    var id: String {get}
+    var action: Action {get}
+    func validate() throws
+}
--- a/LoopKit/Service/RemoteDataService.swift
+++ b/LoopKit/Service/RemoteDataService.swift
@@ -104,11 +104,11 @@ public protocol RemoteDataService: Service {
     func uploadSettingsData(_ stored: [StoredSettings], completion: @escaping (_ result: Result<Bool, Error>) -> Void)
     
     /**
-     Validates a push notification originated from this data service.
+     Fetch a RemoteCommand from a push notification
      - Parameter notification: The push notification dictionary
-     - Returns: Success
+     - Returns: RemoteCommand
      */
-    func validatePushNotificationSource(_ notification: [String: AnyObject]) -> Result<Void, Error>
+    func commandFromPushNotification(_ notification: [String: AnyObject]) async throws -> RemoteCommand
 
 }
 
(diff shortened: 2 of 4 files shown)


LoopKit/Service/Remote/RemoteCommand.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  RemoteCommand.swift
+//  LoopKit
+//
+//  Created by Bill Gestrich on 2/25/23.
+//  Copyright © 2023 LoopKit Authors. All rights reserved.
+//
+
```
gestrich, line 8: This is the shared model that will be returned by plugins to Loop. It opens an API into the plugin where state transitions will be reported during the command enactment process.

LoopKit/Service/RemoteDataService.swift
```diff
@@ -104,11 +104,11 @@ public protocol RemoteDataService: Service {
...
     
     /**
-     Validates a push notification originated from this data service.
+     Fetch a RemoteCommand from a push notification
      - Parameter notification: The push notification dictionary
-     - Returns: Success
+     - Returns: RemoteCommand
      */
-    func validatePushNotificationSource(_ notification: [String: AnyObject]) -> Result<Void, Error>
+    func commandFromPushNotification(_ notification: [String: AnyObject]) async throws -> RemoteCommand
```
ps2, line 111: Nit: The docs say "fetch", but the signature looks like a simple conversion. Would be nice if the caller knows if this is a method that may have network level latencies.

LoopKit/Service/Remote/RemoteCommand.swift
ps2: I keep thinking about this architecture and the questions around it (where do the "command" or "action" models live, and who "owns" them?) and it feels like it's maybe not the right abstraction. 

Just this morning I had an idea that might simplify this.  Ultimately, we're implementing a feature where Loop exposes a remote control interface to service plugins.  The data transfer, serialization, validation, authorization,  etc, that a service plugin needs to implement the functionality should be all responsibilities of the service plugin, and not Loop. We're mostly there, I think, but if Loop isn't going to be managing "commands" and "actions", then why is this a model, and not just a simple interface that Loop exposes to the service plugins?

```
public protocol RemoteCommandHandler {
  func enactBolus(units: Double, initiatedBy: String) async throws
  func scheduleOverride(name: String, duration: TimeInterval, initiatedBy: String) async throws
  func cancelOverride(name: String, initiatedBy: String) async throws
  func addCarbEntry(amountInGrams: Double, absorptionTime: TimeInterval?, foodType: String?, startDate: Date?, initiatedBy: String) async throws
}

extension ServiceDelegate: RemoteCommandHandler { }
```

ps2: This feels more natural to me, and clearly documents exactly what Loop is exposing more succinctly than a new type system of`RemoteCommand`s that can have validations, id, and actions, which have subtypes, are codable, etc.

PR #457, opened 2023-03-06
Format emoji sources, added emjoi names

diggabyte:
* Formatting-only changes to allow future diffs easier to read 
* Added emjoi names for readability
* No changes to food or groupings


ps2:
Thanks!

PR #464, opened 2023-03-20
feat(lantus): adding in lantus

bassrock:
I use Loop to manage and track my insulin/values on my phone in an open loop setup. Because of this I, personally, needed a way to log basal (Lantus) manually.

### DIFF
--- a/LoopKit/InsulinKit/HKQuantitySample+InsulinKit.swift
+++ b/LoopKit/InsulinKit/HKQuantitySample+InsulinKit.swift
@@ -40,6 +40,7 @@ extension HKQuantitySample {
         var metadata: [String: Any] = [
             HKMetadataKeySyncVersion: syncVersion,
             HKMetadataKeySyncIdentifier: syncIdentifier,
+            HKMetadataKeyWasUserEntered: dose.manuallyEntered,
             MetadataKeyHasLoopKitOrigin: true,
             MetadataKeyManuallyEntered: dose.manuallyEntered
         ]
@@ -54,11 +55,6 @@ extension HKQuantitySample {
             metadata[MetadataKeyIsSuspend] = true
 
         case .basal, .tempBasal:
-            // Ignore 0-duration basal entries
-            guard dose.endDate.timeIntervalSince(dose.startDate) > .ulpOfOne else {
-                return nil
-            }
-
             metadata[HKMetadataKeyInsulinDeliveryReason] = HKInsulinDeliveryReason.basal.rawValue
 
             if let basalRate = dose.scheduledBasalRate {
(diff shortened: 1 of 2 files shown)


LoopKit/InsulinKit/HKQuantitySample+InsulinKit.swift
```diff
@@ -40,17 +40,13 @@ extension HKQuantitySample {
...
             HKMetadataKeySyncIdentifier: syncIdentifier,
+            HKMetadataKeyWasUserEntered: dose.manuallyEntered,
             MetadataKeyHasLoopKitOrigin: true,
             MetadataKeyManuallyEntered: dose.manuallyEntered
         ]
         
         switch dose.type {
         case .basal, .tempBasal, .suspend:
-            // Ignore 0-duration basal entries
-            guard dose.endDate.timeIntervalSince(dose.startDate) > .ulpOfOne else {
```
bassrock, line 58: Lantus (basal) is injected at a single time. Technically does not have an end date. I can not find guidance how to log injected basal insulin. For instance, if I take 14units, do you set the start/end date to the same? Or set the end date to 24 hours in the future? 🤷  So because of that removing this guard.

LoopKit/InsulinKit/HKQuantitySample+InsulinKit.swift
```diff
@@ -40,17 +40,13 @@ extension HKQuantitySample {
         var metadata: [String: Any] = [
             HKMetadataKeySyncVersion: syncVersion,
             HKMetadataKeySyncIdentifier: syncIdentifier,
+            HKMetadataKeyWasUserEntered: dose.manuallyEntered,
```
ps2, line 43: I like this change.. Could you submit it as a separate ticket?

LoopKit/InsulinKit/InsulinType.swift
```diff
@@ -64,15 +69,26 @@ public enum InsulinType: Int, Codable, CaseIterable {
...
         switch self {
-        case .afrezza:
+        case .afrezza, .lantus:
             return false
         default:
             return true
         }
     }
+    
+    public var isBasal: Bool {
```
ps2, line 86: Where is this being used? I assume you would *not* want these doses to affect Loop's calculations, right?  Since Loop has a relative view of basal.

ps2 (commented):
This is interesting, and I think maybe it could be accepted, but I'm not sure I fully understand the use case. Loop is an automated delivery system that relies on being able to stop basal insulin delivery to avoid low bgs. How are you using Loop when manually dosing Lantus?

bassrock:
@ps2 Yea, unfortunately I can't get a pump yet, and Loop is the best app I could find that plays so nicely with the Apple ecosystem (and one whos code I liked). So I am using it to manually enter all my doses, submit it to HealthKit & Nightscout and for the IOB estimates.  And sometimes for the carb estimates.

PR #476, opened 2023-04-26
Fix crash caused by invalid ClosedRange in lowerBoundRange and upperBoundRange for mmol/L

### LINKED ISSUE #1864: Glucose Range Picker causes crash using mmol/L and min or max values
**Describe the bug**
When using mmol/L as glucoses units, all instances of GlucoseRangePicker (used to select normal correction range and premeal correction range) share a common bug: when setting the range to the lowest possible value to the lowest possible value OR the highest possible value to the highest possible value the app oftentimes crashes.

**To Reproduce**
1. Use mmol/L as glucose unit
2. Adjust correction range to 4.4-4.4 or 10.0-10.0 in either onboarding or settings (likely chance of crash)
3. Adjust premeal range to 4.4-4.4 or 10.0-10.0 in either onboarding or settings (likely chance of crash)

**Expected behavior**
The app should accept these values

**Phone**
 - Simulator, iOS 16.1

**Loop Version**
 - Current dev workspace (9fd16c2) with the newest LoopKit/LoopKit submodule (7134679)

**Why this happens**
From brief sleuthing, it appears to be due to a rounding error from mg/dL to mmol/L where the minValue is just higher than the maxValue of the range.

bjorkert:
This PR addresses an issue that caused a crash when creating a ClosedRange with invalid bounds when using mmol/L units. Specifically, when using the lowest range (e.g., 4.8 to 4.8), the min value was calculated as 87, while the max value was calculated from 4.8, which equals 86.4. This resulted in an invalid range from 87 to 86.4, causing a crash. The same issue occurred for the highest range (e.g., 10 to 10). This problem did not occur when using mg/dL units.

To fix this issue, this PR ensures that the max value in lowerBoundRange is not less than guardrail.absoluteBounds.lowerBound, and that the min value in upperBoundRange is not greater than guardrail.absoluteBounds.upperBound.

These changes should provide more robust handling of edge cases in the calculation of lowerBoundRange and upperBoundRange when using mmol/L units. The solution has also been tested for mg/dL units to ensure there is no regression.

Resolves https://github.com/LoopKit/Loop/issues/1864


marionbarker:
I just tested this with most recent dev (commit 67a4b1f) and it still hangs at 
lines of LoopWorkspace/LoopKit/LoopKitUI/Views/GlucoseRangePicker.swift that are modified by this PR.

Test 1: test correction range to 4.8 to 4.8 - hangs
* I modified the code as shown in first portion of this PR - successfully set to 4.8 to 4.8 correction range.

Test 2: test correction range to 10.0 to 10.0 - hangs
* I modified the code as shown in second portion of this PR - successfully set to 10.0 to 10.0 correction range.

ps2:
@bjorkert Thanks for working on this. I believe the problem is actually being caused earlier, where outside of guardrails values are allowed to be selected. I've created a fix here: https://github.com/LoopKit/LoopKit/pull/527, which fixes the issue (and the crash) earlier.

PR #478, opened 2023-05-01
Disable missed meal detection if there are user-entered or calibration BGs

novalegra:
This makes `isCalibrated` accessible from Loop. I also updated it to be a function since the runtime isn't constant-time

### DIFF
--- a/LoopKit/GlucoseKit/GlucoseMath.swift
+++ b/LoopKit/GlucoseKit/GlucoseMath.swift
@@ -48,7 +48,7 @@ extension BidirectionalCollection where Element: GlucoseSampleValue, Index == In
 
     /// Whether the collection contains no calibration entries
     /// Runtime: O(n)
-    var isCalibrated: Bool {
+    public func isCalibrated() -> Bool {
         return filter({ $0.isDisplayOnly }).count == 0
     }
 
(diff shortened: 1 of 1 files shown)


LoopKit/GlucoseKit/GlucoseMath.swift
```diff
@@ -48,7 +48,7 @@ extension BidirectionalCollection where Element: GlucoseSampleValue, Index == In
 
     /// Whether the collection contains no calibration entries
     /// Runtime: O(n)
-    var isCalibrated: Bool {
+    public func isCalibrated() -> Bool {
```
ps2, line 51: maybe `containsCalibration` or `hasCalibration`? I wouldn't say the trace "is calibrated", just because it contains an entry that is a calibration.

ps2: And why is it turning into a function?  It takes no args and has no side effects, so it could still remain a var?

novalegra: Agreed on the naming! I made it a function since I normally think of variables as having constant-time access; happy to stick to a variable since it's a change that's not related to the core of the PR

novalegra: I'm honestly a little confused by the comment ("contains no calibration entries") versus the name ("isCalibrated"). It seems like this function is seeing if we have no points that, if there had been a calibration, were shifted for visual consistency, so I would expect the name to be something like "isNotCalibrated". Which interpretation would be correct?

ps2: Yeah, the function name is not great; probably `containsCalibrations()` is better.

ps2: Or negated, `containsNoCalibrations()`

ps2:
Landed here: https://github.com/LoopKit/LoopKit/commit/08e028e64441e18d2a9708fd0a3b6902fd26f716

PR #482, opened 2023-05-19
make InstructionList usable with markdown content, to display links, bo…

dabear:
@marionbarker requested for LibreTransmitter that the GuidePage for setting up new Libre sensors contain a clickable link to the readme file. 

This implements the necessary changes to the InstructionList in LoopKit to make this possible. Note that this implementation toggles "markdown mode" for all the instructions within an InstructionList rather than individually. 

The default is off, due to a couple of reasons:
* Potentially there could be callers to InstructionList containing text that would be interpreted as markdown where it shouldn't (backward compatibility )
* According to Apple documentation, when using LocalizedStringKey any localizations done automatically by swiftui will be turned off. Therefore, when enabling markdown mode, callers need to localize the text before passing it into InstructionList. (I.e. they must use NSLocalizedString) 

iPhone_14_Pro_Max

If this passes, I will followup with a PR for LibreTransmitter with the following changes (this is just for reference):
```diff
index e368540..a8a8475 100644
--- a/LibreTransmitterUI/Views/Setup/ModeSelectionView.swift
+++ b/LibreTransmitterUI/Views/Setup/ModeSelectionView.swift
@@ -58,6 +58,9 @@ struct ModeSelectionView: View {
         }// .accentColor(.red)
     }

+
+
+
     var body : some View {
         GuidePage(content: {
             VStack {
@@ -67,9 +70,9 @@ struct ModeSelectionView: View {
                     InstructionList(instructions: [
                         LocalizedString("Sensor should be activated and fully warmed up", comment: "Label text for step 1 of connection setup"),
                         LocalizedString("Select the type of setup you want.", comment: "Label text for step 2 of connection setup"),
-                        LocalizedString("Not all sensor types can be supported, see readme.md", comment: "Label text for step 3 of connection setup"),
-                        LocalizedString("Fair warning: The sensor will be not be using the manufacturer's algorithm, and some safety mitigations present in the manufacturers algorithm might be missing when you use this.", comment: "Label text for step 4 of connection setup")
-                    ])
+                        LocalizedString("Not all sensor types can be supported, see [readme.md](https://github.com/LoopKit/LibreTransmitter/blob/main/readme.md)", comment: "Label text for step 3 of connection setup"),
+                        LocalizedString("Fair warning: The sensor will be not be using the manufacturer's algorithm, and some safety mitigations present in the manufacturers algorithm might be missing when you use this.", comment: "Label text for step 4 of connection setup"),
+                    ], markdownEnabled: true)
                 }

             }
```

### DIFF
--- a/LoopKitUI/Views/InstructionList.swift
+++ b/LoopKitUI/Views/InstructionList.swift
@@ -17,18 +17,24 @@ public struct InstructionList: View {
     let startingIndex: Int
     let instructionColor: Color
     
+    // Setting this to true will also disable swiftui's localization,
+    // so the instruction should in those cases be pre-localized in those cases
+    public var markdownEnabled: Bool
+    
     @Environment(\.isEnabled) var isEnabled
     
-    public init(instructions: [String], startingIndex: Int = 1, instructionColor: Color = .primary) {
+    public init(instructions: [String], startingIndex: Int = 1, instructionColor: Color = .primary, markdownEnabled : Bool = false) {
         self.instructions = instructions.map { Instruction(text: $0, subtext: nil) }
         self.startingIndex = startingIndex
         self.instructionColor = instructionColor
+        self.markdownEnabled = markdownEnabled
     }
     
-    public init(instructions: [(String, String)], startingIndex: Int = 1, instructionColor: Color = .primary) {
+    public init(instructions: [(String, String)], startingIndex: Int = 1, instructionColor: Color = .primary, markdownEnabled : Bool = false) {
         self.instructions = instructions.map { Instruction(text: $0.0, subtext: $0.1) }
         self.startingIndex = startingIndex
         self.instructionColor = instructionColor
+        self.markdownEnabled = markdownEnabled
     }
     
     public var body: some View {
@@ -42,7 +48,7 @@ public struct InstructionList: View {
                         .foregroundColor(Color.white)
                         .font(.caption)
                         .accessibility(label: Text("\(index+1), ")) // Adds a pause after the number
-                    instructionView(instructions[index])
+                    instructionView(instructions[index], markdownEnabled: markdownEnabled)
                 }
                 .accessibilityElement(children: .combine)
             }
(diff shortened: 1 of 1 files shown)


dabear:
this:
```swift
if markdownEnabled {
                Text(LocalizedStringKey(instruction.text))
                    .fixedSize(horizontal: false, vertical: true)
            } else {
                Text(instruction.text)
                    .fixedSize(horizontal: false, vertical: true)
            }
```
could probably me more compact, but decided not to do so for readability concerns

dabear:
@ps2 this is ready for review

ps2:
1) It would be nice to put the list of supported sensors into the onboarding screens. That works offline then, too, and doesn't take them out of the app.
2) We should be able to detect the variant of sensor by a scan, right? And tell them if it's supported or not.

LoopKitUI/Views/InstructionList.swift
```diff
@@ -42,23 +48,37 @@ public struct InstructionList: View {
...
-    private func instructionView(_ instruction: Instruction) -> some View {
+    private func instructionView(_ instruction: Instruction, markdownEnabled: Bool = false) -> some View {
         VStack(alignment: .leading, spacing: 2) {
-            Text(instruction.text)
-                .fixedSize(horizontal: false, vertical: true)
-            if let subtext = instruction.subtext {
-                Text(subtext)
+            
+            if markdownEnabled {
+                Text(LocalizedStringKey(instruction.text))
```
ps2, line 63: LocalizedStringKey looks up strings for localization in the current module's bundle; in this case, LoopKit.  The localization lookup for string that originate in a different module should be handled in that module.

dabear: Was this informational only, or did you want me to do any change there?

dabear:
> * It would be nice to put the list of supported sensors into the onboarding screens. That works offline then, too, and doesn't take them out of the app.
> * We should be able to detect the variant of sensor by a scan, right? And tell them if it's supported or not.

Putting a complete list of supported sensors is kind of hard since there are multiple versions and "security generations" within a version of a sensor . However I could change the text to "most libre 1 and libre 2 sensors (except us and Canada) are supported" if that makes more sense?

We can and do detect if the tag is supported in the code. However the feedback to the end user possibly could be improved.
 
Better feedback to the end user is out of scope for this PR but could be handled in a follow-up Pr to the libretransmitter repo

Let me know if you want me to close this PR or if we can make change

dabear:
still awaiting an answer on this :)

LoopKitUI/Views/InstructionList.swift
ps2: This will not work for localization.

ps2: One option might be to change this class to accept `LocalizedStringKey` for instruction text & subtext, instead of `String`.  The localization needs to happen in the caller's framework, though.

ps2:
I'm not a fan of linking out like this; it's disruptive to the process, and they have to find their way back to the app.

ps2:
> when using LocalizedStringKey any localizations done automatically by swiftui will be turned off

I think maybe this was a mis-read.  When passing a `String` variable to initializers like `Text(...)`, then localization lookup is not performed. `LocalizedStringKey(stringVar)` will force there to be a lookup on the contents of stringVar, but we shouldn't be doing lookups on strings provided by other frameworks, as those frameworks should be doing the lookups against their bundles.

LoopKitUI/Views/InstructionList.swift
ps2: Scratch that idea of accepting a LocalizedStringKey. That would still trigger an app bundle translation lookup when it's used in the `Text` view.

dabear:
> > when using LocalizedStringKey any localizations done automatically by swiftui will be turned off
> 
> I think maybe this was a mis-read. When passing a `String` variable to initializers like `Text(...)`, then localization lookup is not performed. `LocalizedStringKey(stringVar)` will force there to be a lookup on the contents of stringVar, but we shouldn't be doing lookups on strings provided by other frameworks, as those frameworks should be doing the lookups against their bundles.

Using LocalizedStringKey seemed like the best/easiest approach to get to that Text initializer that supports markdown. 

 I see now that I could replace that with the following initializer of the Text class instead, since that also supports markdown: 
```public init(_ attributedContent: AttributedString)```

That would mean a change to the code to fascilitate this change:
```
public struct InstructionList: View {
    struct Instruction {
        let text: AttributedString
        let subtext: AttributedString?
    }
//(...)
 ```
 However, I see that (as I Interpret your previous answer)  having markdown and/or support for links inside a instructionlist in a guidepage is not a wanted feature, so I'll just close this ticket for now

PR #486, opened 2023-05-26
Profile switching

bjorkert:
In collaboration with Marion Barker, who has provided invaluable feedback, performed real-world testing, and offered ongoing encouragement, I have developed a Profile Switcher for Loop. This feature empowers users to effortlessly shift between settings that cater to differing insulin needs, simply by switching between defined profiles, such as school and weekend profiles.

The "Profiles" feature resides under "Therapy Settings" in the "Configuration" section. Each profile comprises a subset of the therapy settings, including:
- Glucose target range
- Carb ratio
- Basal rates
- Insulin sensitivity settings

We've deliberately chosen to limit the range of settings stored in a profile, excluding aspects such as overrides and pre-meal range. This feature isn't intended as a backup tool, and we've aimed to avoid cluttering profiles with too many settings that could necessitate manual adjustments across all profiles when a universally applicable change is required. Nonetheless, I appreciate that this selection might not perfectly fit all users' needs, while my hope is that it aligns with what the majority would prefer.

Profiles are stored as JSON files in a dedicated folder within the app's private sandbox directory. Users can create a profile using their current settings, and prior to loading, each profile is presented on a read-only screen for review.

When a profile is loaded, the settings are first verified against guardrails, pump increments, and max basal settings. Once validated, the basal rates are synchronized with the pump. Successful synchronization is followed by an update of the therapy settings.

All features are currently functioning as intended. The ordering of profiles isn't implemented yet, and user-defined sorting is a desirable enhancement.

A frequent request related to profiles is scheduling of profiles, it is not implemented in this PR.

There are profile functionality also in other related systems like Nightscout but as it is now only the default profile is updated with the values that are loaded into the therapy settings.

Screenshots
Profiles entry
Profile list
Create profile
Preview prpfile
Delete profile
Erorr loading profile
Load/Delete

### DIFF
--- a/LoopKit.xcodeproj/project.pbxproj
+++ b/LoopKit.xcodeproj/project.pbxproj
@@ -852,6 +852,12 @@
 		C1FAEC1D264AD6B400A3250B /* DeviceStatusBadge.swift in Sources */ = {isa = PBXBuildFile; fileRef = C1FAEC1C264AD6B400A3250B /* DeviceStatusBadge.swift */; };
 		C1FAEC1F264AE12700A3250B /* UIImage.swift in Sources */ = {isa = PBXBuildFile; fileRef = B47ECF8725DC20810024A54D /* UIImage.swift */; };
 		C1FAEC21264AEEA300A3250B /* UIImage.swift in Sources */ = {isa = PBXBuildFile; fileRef = C1FAEC20264AEEA300A3250B /* UIImage.swift */; };
+		DD508E062A17712D00EEF8FD /* ProfileViewModel+FileManagement.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD508E052A17712D00EEF8FD /* ProfileViewModel+FileManagement.swift */; };
+		DD508E082A17763700EEF8FD /* NewProfileEditor.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD508E072A17763700EEF8FD /* NewProfileEditor.swift */; };
+		DD7122612A1D2AA4004FE653 /* ProfilePreviewView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD7122602A1D2AA4004FE653 /* ProfilePreviewView.swift */; };
+		DDAF746729F422C600719F0A /* ProfileView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DDAF746629F422C600719F0A /* ProfileView.swift */; };
+		DDAF746929F4234000719F0A /* ProfileViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = DDAF746829F4234000719F0A /* ProfileViewModel.swift */; };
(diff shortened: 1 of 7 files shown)


ps2:
This was discussed elsewhere (in zulip, I think), but the main thing that needs to be addressed there is that rather than overwriting Loop's main profile each time, Loop should be updated to be "multi profile" aware in a fuller sense, so that the profile set can be uploaded to backend services, and switches from one profile to another are recorded semantically as switching a profile, and not overwriting a profile.

bjorkert:
Profiles has been updated with a few new features:
- Indication of the currently active profile with a checkmark.
- Manual Profile Sorting with drag and drop.
- After modifying therapy settings, Profiles will now suggest updating your profile. Alternatively, save these new settings as a separate profile.
- Profile Renaming

profiles_v2

bjorkert:
> This was discussed elsewhere (in zulip, I think), but the main thing that needs to be addressed there is that rather than overwriting Loop's main profile each time, Loop should be updated to be "multi profile" aware in a fuller sense, so that the profile set can be uploaded to backend services, and switches from one profile to another are recorded semantically as switching a profile, and not overwriting a profile.

Having them uploaded would also make it possible to retrieve them during the onboarding, showing the current profile in Nightscout is a requested feature. In the latest version, Loop is made aware of the latest Loaded (and therefore considered the current) profile. 

How do you see the workflow for uploading "multi profile". For example: Should there be "Default" until a profile is created, or all profiles are removed. When saving a profile, it should trigger an update. Switching profile should set the current one as active in Nightscout. When having a profile active, making changes in therapy settings without saving them to the/a profile, what would that mean? I would very much like some reasoning about how you think about things like these.

ps2:
Good questions.  So yes, to minimize changes to existing UX in onboarding and such, there should be a "default".  

The upload should work how you describe; when a profile is changed, it needs to be uploaded, and when the currently active one is changed, that needs to be uploaded as well. To do this, though, you'll need the below change.

The biggest change is rather than storing profiles in files on the phone, the core `StoredSettings` data model needs to be updated to include the full set of profiles, and which one is active.  This gives us a full history of what settings where at any point in time, and will be included in an export of critical logs, etc. And this data model change will propagate through the RemoteDataService parts of Loop, so any service (not just nightscout) could store this data as well.

ps2:
After the storage is updated and the data is piped to remote data services, then the last part is probably making sure the UX is good for onboarding and settings editing.  I don't think that a user should have to edit settings, and *then* decide if they want to update the saved profile.  When they are using a non-default profile, and they edit it, that should be clear user intent.  It's important that they know what they are doing, though, so when editing settings and using a non-default profile, we should be clear about that in the UI.

There were also some user requests about making more visible which profile is currently being used on the main status page.

bjorkert:
Thanks for the feedback! Using StoredSettings and storing changes to the current profile makes things clearer, therapy settings should perhaps indicate that there is an active profile as well.

Regarding the UI for active profiles in the main view, do you already have a design approach for how this should be highlighted?

I'll look in to how StoredSettings is structured and next step would probably be a suggestion of how a multi profile StoredSettings could look like.

PR #488, opened 2023-06-12
Remote PR Set #3

gestrich:
This is part of a collection of foundational PRs that need to merge at the same time.

[Loop](https://github.com/LoopKit/Loop/pull/2009)
[LoopKit](https://github.com/LoopKit/LoopKit/pull/488)
[NightscoutService](https://github.com/ps2/NightscoutService/pull/41)
[TidepoolService](https://github.com/LoopKit/TidepoolService/pull/13)


LoopKit/Service/RemoteDataService.swift
```diff
@@ -104,11 +104,13 @@ public protocol RemoteDataService: Service {
     func uploadSettingsData(_ stored: [StoredSettings], completion: @escaping (_ result: Result<Bool, Error>) -> Void)
     
     /**
-     Fetch a RemoteCommand from a push notification
+     Handle a push notification
      - Parameter notification: The push notification dictionary
-     - Returns: RemoteCommand
      */
```
gestrich, line 109: REMOTE 2.0: WILL REMOVE PRE-MERGE

LoopKit/Service/RemoteDataService.swift
```diff
@@ -104,11 +104,13 @@ public protocol RemoteDataService: Service {
...
-    func commandFromPushNotification(_ notification: [String: AnyObject]) async throws -> RemoteCommand
+    func handleRemoteNotification(_ notification: [String: AnyObject]) async throws
+
```
gestrich: REMOTE 2.0: WILL REMOVE PRE-MERGE

LoopKit/Service/Remote/RemoteActionDelegate.swift
```diff
@@ -0,0 +1,18 @@
...
+//  Copyright © 2023 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+public protocol RemoteActionDelegate: AnyObject {
+    func handleRemoteOverride(name: String, durationTime: TimeInterval?, remoteAddress: String) async throws
+    func handleRemoteOverrideCancel() async throws
+    func handleRemoteCarb(amountInGrams: Double, absorptionTime: TimeInterval?, foodType: String?, startDate: Date?) async throws
+    func handleRemoteBolus(amountInUnits: Double) async throws
```
gestrich: REMOTE 2.0: WILL REMOVE PRE-MERGE

LoopKit/Service/Remote/RemoteActionDelegate.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  RemoteActionDelegate.swift
+//  LoopKit
+//
+//  Created by Bill Gestrich on 3/19/23.
+//  Copyright © 2023 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
```
gestrich, line 10: The `RemoteActionDelegate` is conformed to both by the ServiceManager and DeviceDataManager to propogate actions to where they are delivered.

I'm using "handleRemote" in the method names to help disambiguate from existing APIs in DeviceDataManaager, like "enactBolus". I can drop the prefix if desired though.

MockKit/MockService.swift
```diff
@@ -168,12 +168,9 @@ extension MockService: RemoteDataService {
...
-    public func commandFromPushNotification(_ notification: [String: AnyObject]) async throws -> RemoteCommand {
-        
-        enum MockServicePushNotificationError: LocalizedError {
-            case remoteCommandsNotSupported
-        }
-        
-        throw MockServicePushNotificationError.remoteCommandsNotSupported
+    public func handleRemoteNotification(_ notification: [String: AnyObject]) async throws {
+    }
+
```
gestrich: REMOTE 2.0: WILL REMOVE PRE-MERGE

LoopKit/Service/RemoteDataService.swift
ps2: nit: "handle" is one of those non-descriptive verbs, like "process".  Every function "handles" something or "processes" something.  Maybe something like `remoteNotificationWasReceived(_ notification ...)`

LoopKit/Service/Remote/RemoteActionDelegate.swift
ps2: Again, "handle" is very vague. This API is exposing dosing functionality and settings change functionality. So it would be more descriptive to have the verbs describe actually enacting or delivering.  I think it's fine to keep the "remote" in there, as that's the main foreseeable use for this API, but an API like this could theoretically be used by Services for other things, like maybe they have some kind of other trigger to change dosing, like a timer, or other notification.

gestrich: Done but let me know if feedback on the names below.

LoopKit/Service/RemoteDataService.swift
gestrich: Done

LoopKit/Service/Remote/RemoteActionDelegate.swift
```diff
@@ -0,0 +1,16 @@
...
+public protocol RemoteActionDelegate: AnyObject {
+    func updateRemoteOverride(name: String, durationTime: TimeInterval?, remoteAddress: String) async throws
```
ps2: Does this enact an override, or just "update" it? I don't see an "enact" version.

ps2 (approved):
LGTM!

LoopKit/Service/Remote/RemoteActionDelegate.swift
gestrich: Yeah enact probably works better. This is for either starting a new override _or_ updating an existing - no distinction between them here.

I pushed an update for the name.

PR #498, opened 2023-07-21
Remove Loop Dependencies from Charts

gestrich:
In order to remove Caregiver's dependency on the Loop repo, I need to move certain charts from Loop to LoopKit. This PR will not move those charts yet but instead updates some chart APIs in prep for that.

* Use `ChartColorPalette` for chart colors: Charts use UIColor extension methods, like `insulinTintColor` which are not available in LoopKit. `ChartColorPalette` is available to LoopKit so this switches to using that. Additionally, I added `carbTint` to `ChartColorPalette` which was missing.
* This needs to merge with https://github.com/LoopKit/Loop/pull/2030

### DIFF
--- a/LoopKitUI/Extensions/Environment+Colors.swift
+++ b/LoopKitUI/Extensions/Environment+Colors.swift
@@ -110,7 +110,8 @@ private struct ChartColorPaletteKey: EnvironmentKey {
                                                                    axisLabel: .secondaryLabel,
                                                                    grid: .systemGray3,
                                                                    glucoseTint: .systemTeal,
-                                                                   insulinTint: .orange)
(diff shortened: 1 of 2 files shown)


LoopKitUI/Extensions/Environment+Colors.swift
```diff
@@ -110,7 +110,8 @@ private struct ChartColorPaletteKey: EnvironmentKey {
                                                                    axisLabel: .secondaryLabel,
```
gestrich, line 110: These colors appear to be fallbacks for the `EnvironmentKey` when there is no value specified. systemGreen seems right as that is the same fallback used in Loop/LoopUI/Extensions/UIColor.swift

PR #508, opened 2023-08-24
Clarify that button clears credentials

marionbarker:
A Facebook user was confused by this button name when trying to add Nightscout as a Remote CGM. 

They had an invalid URL in their Nightscout account and did not understand that they needed to hit "Delete" in order to be able to enter a new URL and API_SECRET.

I believe changing this button from "Delete" to "Clear Credentials" would work for all services. It does what I expect for Nightscout.

clear-creditials-button-text


ps2:
The label on the button should describe the removal of the service, which means all aspects of this service, including credentials, or other configuration, so I think this change is not semantically correct.  If we wanted to have a way to keep the service, and change credentials (maybe what the original user issue was about), then we could either 1) add a way to edit credentials, or 2) communicate that to change credentials, they must remove and re-add the service.

PR #519, opened 2023-12-17
Support LoopKit Package in Watch targets

gestrich:
The LoopKit Swift Package target is almost Watch compatible, except for 2 methods. This prevents LoopKit from being imported into watch apps, which is going to be helpful for the upcoming Loop Caregiver watch app and any other Watch apps.


ps2 (approved):
Thanks!

PR #525, opened 2024-02-06
Add Simulator Settings Row to DIY version for pump and CGM simulators

marionbarker:
The simulators for pump and CGM do not have an obvious way to delete them so an actual pump and CGM can be added and the time required for the long press is very long.

Based on conversations from my initial attempt, that commit was reverted with a new approach used.

See later comments for continued conversation.

### DIFF
--- a/MockKitUI/MockCGMManager+UI.swift
+++ b/MockKitUI/MockCGMManager+UI.swift
@@ -15,7 +15,6 @@ import MockKit
 
 
 extension MockCGMManager: CGMManagerUI {
-
     fileprivate var appName: String {
         return Bundle.main.object(forInfoDictionaryKey: "CFBundleDisplayName") as! String
     }
@@ -29,7 +28,7 @@ extension MockCGMManager: CGMManagerUI {
     }
 
     public func settingsViewController(bluetoothProvider: BluetoothProvider, displayGlucosePreference: DisplayGlucosePreference, colorPalette: LoopUIColorPalette, allowDebugFeatures: Bool) -> CGMManagerViewController {
-        let settings = MockCGMManagerSettingsView(cgmManager: self, displayGlucosePreference: displayGlucosePreference, appName: appName)
+        let settings = MockCGMManagerSettingsView(cgmManager: self, displayGlucosePreference: displayGlucosePreference, appName: appName, allowDebugFeatures: allowDebugFeatures)
         let hostingController = DismissibleHostingController(content: settings, isModalInPresentation: false, colorPalette: colorPalette)
         hostingController.navigationItem.backButtonDisplayMode = .generic
         let nav = CGMManagerSettingsNavigationViewController(rootViewController: hostingController)
--- a/MockKitUI/MockCGMManager+UI.swift
+++ b/MockKitUI/MockCGMManager+UI.swift
@@ -15,6 +15,7 @@ import MockKit
 
 
 extension MockCGMManager: CGMManagerUI {
+
     fileprivate var appName: String {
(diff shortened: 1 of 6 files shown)


scifantastic:
Thanks!  There are a few requirements for this, if you're up for it:
* We can add easy access to simulator configuration for DIY, but it needs to be behind a feature flag.  Could be something like APP_DEMO_MODE, and when you're in demo mode, the simulator configuration is hard to get to (i.e., how it is in dev now).
* I think we can add back a top level "remove simulator" command, but if, and only if, there are other CGM plugins configured.
* When in easy access mode (non-demo), I think it'd be more congruent with the design to have the configuration page accessible like other configuration items on devices; like the "Device Details" page, rather than a "Press and hold for more options" text splatted at the top.  The title could be "Configure Simulator". Again, this would only be available in non-demo mode.  In demo mode, the simulator configuration needs to be near impossible to accidentally activate.

marionbarker:
Happy to work with you on this. How about something like this instead? (This is from @bjorkert - he and I often work together.)

Alternative-MockCGM-with-Settings-Button-2024-02-06 21 14 57

scifantastic:
Yes, that would be fine.

ps2:
[reposting with correct account]

Thanks! There are a few requirements for this, if you're up for it:

We can add easy access to simulator configuration for DIY, but it needs to be behind a feature flag. Could be something like APP_DEMO_MODE, and when you're in demo mode, the simulator configuration is hard to get to (i.e., how it is in dev now).
I think we can add back a top level "remove simulator" command, but if, and only if, there are other CGM/pump plugins configured.
When in easy access mode (non-demo), I think it'd be more congruent with the design to have the configuration page accessible like other configuration items on devices; like the "Device Details" page, rather than a "Press and hold for more options" text splatted at the top. The title could be "Configure Simulator". Again, this would only be available in non-demo mode. In demo mode, the simulator configuration needs to be near impossible to accidentally activate.

marionbarker:
In order to get a new Feature Flag into LoopKit, we think we would have to modify the protocol for all pump and cgm managers.

It looks like the existing DEBUG_FEATURES_ENABLED / allowDebugFeatures variables can be used because that is already in the pump and cgm managers. We reviewed the places where it is applied and it looks reasonable to configure the DIY code to have this enabled. If this is incorrect, then guidance is requested for what change should be made.

This updated version builds to the normal, long-press required version of the Simulators unless the DEBUG_FEATURES_ENABLED is uncommented from the LoopConfigOverride.xcconfig file.

loopkit-pr-525

marionbarker:
I searched for places where having DEBUG_FEATURES_ENABLED enabled for DIY would change the app. (Note the user would need to know how to enable this easter egg):

This is only one I found. (Other items are handled by longPress regardless of this flag).

* https://github.com/LoopKit/Loop/blob/dev/Loop/View%20Controllers/StatusTableViewController.swift#L1844-L1926

> // Opens the debug menu if you rotate the phone 6 times (or back & forth 3 times), each rotation within 2 secs.

This is the view on my test phone:

DebugMenu-after-rotation

I tried all the buttons and they appear to work as expected.

~~Furthermore - I was able to solve the Critical Log Export problem described in [Export Critical Log fails](https://github.com/LoopKit/Loop/issues/2081) by deleting the Export Folder. I could then Export Critical Logs without a problem.~~
Edited to add - I was incorrect - it just took longer for the Error to appear when trying to export the critical logs. This is still a problem on a phone that has main built on top of dev, even after returning to dev.

ps2:
Thanks!

PR #527, opened 2024-02-19
Fix issue where glucose values outside of guardrails could be selected.

ps2:
Also fixes the crash described here: https://github.com/LoopKit/Loop/issues/1864


marionbarker:
This does "fix" the crash but it also changes the minimum and maximum ranges available in mmol/L.
After applying this patch, the minimum and maximum range is reduced
* before, the picker allowed 4.8 to 10.0 but would not allow single value correction settings that were at the minumum or maximum, i.e., 4.8-4.8 or 10.0-10.0
* after, the picker allowed range is limited in 4.9 to 9.9 mmol/L
* note the Glucose Safety lower limit remains at 3.7 mmol/L

| mmol/L | mg/dL |
|---|---|
| 3.7 | 66.7 |
| 4.8 | 86.4 |
| 4.9 | 88.2 |
| 9.9 | 178.2 |
| 10.0 | 180.0|

ps2:
Yes, some previous values that appeared to be "acceptable" were actually below/above the guardrail values, due to rounding, causing this crash. As discussed in other tickets, making the guardrails configurable at a build level would be the way to allow the range to be extended.

PR #532, opened 2024-04-22
Allow critical event log to be exported with mixed 3.2.x and 3.3 formats

marionbarker:
When switching from version 3.3 or later code to an earlier version such as 3.2.x code, the Critical Event Log can no longer be exported (from with 3.2.x or 3.3 code). See [Loop Issue 2081: Export Critical Log fails](https://github.com/LoopKit/Loop/issues/2081) 

This work-around allows the export to continue past errors.
* The user will still not be able to export the Critical Event Log while running version 3.2.3 if they switched to 3.3 and back again
* The user will be able to export the Critical Event Log while running version 3.3
   * The caveat is that the DosingDecision.json files reports "Failure to encode" for all entries that occurred while user was running 3.2.x after running 3.3.

## Test 1
* test phone has been running 3.3 (commit ed8d7a0)
   * demonstrate success with Export Critical Event Log
* install 3.2.3 onto phone
   * demonstrate failure with Export Critical Event Log
* Add 2 carbs entries while on 3.2.3 version
* Return to 3.3
   * demonstrate failure with Export Critical Event Log
* Build with new version from this PR
   * demonstrate success with Export Critical Event Log
 
Examine the Critical Event Log for the current day:
* The Carb.json file contains all the entered carbs
* The DosingDecisions.json is the only json file with the "Failure to encode" indications:

```
% find . -name "*.json" -exec grep -Hn "Failed to encode" {} \;
./dev-after-fix-Export-20240422T131220Z/20240422T131220Z/DosingDecisions.json:115:{"error": "Failed to encode item due to: The data couldn’t be read because it is missing."},
./dev-after-fix-Export-20240422T131220Z/20240422T131220Z/DosingDecisions.json:116:{"error": "Failed to encode item due to: The data couldn’t be read because it is missing."},
```

### DIFF
--- a/LoopKit/JSONStreamEncoder.swift
+++ b/LoopKit/JSONStreamEncoder.swift
@@ -45,9 +45,19 @@ public class JSONStreamEncoder {
         }
 
         for value in values {
-            try stream.write(encoded ? ",\n" : "[\n")
-            try stream.write(try Self.encoder.encode(value))
-            encoded = true
+            do {
+                let encodedData = try Self.encoder.encode(value)
+                try stream.write(encoded ? ",\n" : "[\n")
+                try stream.write(encodedData)
+                encoded = true
+            } catch {
+                // Log the error and insert a placeholder or error message in the JSON
+                print("Failed to encode value: \(value) with error: \(error.localizedDescription)")
+                let errorInfo = "{\"error\": \"Failed to encode item due to: \(error.localizedDescription)\"}"
+                try stream.write(encoded ? ",\n" : "[\n")
+                try stream.write(errorInfo.data(using: .utf8)!)
+                encoded = true
+            }
         }
     }
 
--- a/LoopKit/JSONStreamEncoder.swift
+++ b/LoopKit/JSONStreamEncoder.swift
@@ -53,7 +53,7 @@ public class JSONStreamEncoder {
             } catch {
                 // Log the error and insert a placeholder or error message in the JSON
                 print("Failed to encode value: \(value) with error: \(error.localizedDescription)")
-                let errorInfo = "{\"error\": \"Failed to encode item due to: \(error.localizedDescription)\"}"
+                let errorInfo = "{\"error\": \"Failed to encode value: \(value) due to: \(error.localizedDescription)\"}"
                 try stream.write(encoded ? ",\n" : "[\n")
                 try stream.write(errorInfo.data(using: .utf8)!)
                 encoded = true


marionbarker:
@ps2 Please respond. This solves the problem for test phones that have had main built over dev.

LoopKit/JSONStreamEncoder.swift
```diff
@@ -45,9 +45,19 @@ public class JSONStreamEncoder {
...
-            encoded = true
+            do {
+                let encodedData = try Self.encoder.encode(value)
+                try stream.write(encoded ? ",\n" : "[\n")
+                try stream.write(encodedData)
+                encoded = true
+            } catch {
+                // Log the error and insert a placeholder or error message in the JSON
+                print("Failed to encode value: \(value) with error: \(error.localizedDescription)")
+                let errorInfo = "{\"error\": \"Failed to encode item due to: \(error.localizedDescription)\"}"
```
ps2: Can we add the value here?  I'm wondering what was actually failing?

marionbarker: I have tagged @bjorkert to make sure he sees this.

This bug has been present since before Oct 2023. 
No one is aware that they have a problem until they need to get help and realize they cannot issue a critical event log. Most people don't build dev and then return to main, but it can happen. The error is reproducible if you build main (3.2.x) onto a phone that previously had dev (3.3.0) and have loop run a couple of cycles and store data. Then you can no longer issue the critical event log with main or if you build dev again.

IMO, this work-around is good enough and can be patched later if a better solution is found. I have tested that it works on test phones.

marionbarker:
I will test this modification and report back.

marionbarker:
Test Phone:
* iPhone SE, iOS 17.5.1
* previously has been loaded with both dev and main on this phone

Test:
* Install 3.2.3 from TestFlight, confirm that when trying to export Critical Event Logs, the app reports `Error Exporting Logs`
* Install dev, commit c4b4588 onto the phone, confirm same error when trying to export Critical Event Logs
* Switch LoopKit to using loopandlearn:critical-log-work-around including the latest commit, now able to export Critical Event Logs successfully

Xcode Debug Log with new test:

```
Failed to encode value: <LoopKit.DosingDecisionObject: 0x302d8a8f0> (entity: DosingDecisionObject; id: 0x9b85941413b2fd92 <x-coredata://390582E7-5AD6-448A-AED1-5535B012995A/DosingDecisionObject/p58414>; data: {
    data = "{length = 50, bytes = 0x62706c69 73743030 de010203 04050607 ... 73746f72 6963616c }";
    date = "2024-05-26 13:46:29 +0000";
    modificationCounter = 1092211;
}) with error: The data couldn’t be read because it is missing.
Failed to encode value: <LoopKit.DosingDecisionObject: 0x302d8af80> (entity: DosingDecisionObject; id: 0x9b8594141392fd92 <x-coredata://390582E7-5AD6-448A-AED1-5535B012995A/DosingDecisionObject/p58415>; data: {
    data = "{length = 50, bytes = 0x62706c69 73743030 de000100 02000300 ... 3f004100 44004500 }";
    date = "2024-05-26 13:46:30 +0000";
    modificationCounter = 1092212;
}) with error: The data couldn’t be read because it is missing.
Type: stdio
```

Critical Event Log results will be in next comment.

marionbarker:
The error is found in the DosingDecision.json for today's date.

Both the full zip and the DosingDecison.json for today are included below.

[dev-after-fix-2-Export-20240526T140849Z.zip](https://github.com/LoopKit/LoopKit/files/15447755/dev-after-fix-2-Export-20240526T140849Z.zip)

[DosingDecisions.json](https://github.com/LoopKit/LoopKit/files/15447756/DosingDecisions.json)

PR #533, opened 2024-06-03
feature: Add Dana(RS-v3 & -i) configurations

bastiaanv:
The PR adds the DanaRS-v3 and Dana-i pump configurations to the MockKit.

The max basal of the Dana's is 16U/hr
The max bolus of the Dana's is 80U

### DIFF
--- a/LoopKit/InsulinKit/DoseStore.swift
+++ b/LoopKit/InsulinKit/DoseStore.swift
@@ -684,6 +684,18 @@ extension DoseStore {
             return
         }
 
+
+        for event in events {
+            if let dose = event.dose {
+//                if let lastReconciliation, !dose.isMutable, dose.endDate > lastReconciliation {
+//                    // Doses with an endDate after lastReconciliation should still be considered ongoing/mutable, as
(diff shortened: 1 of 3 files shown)


marionbarker:
## Test

I applied this patch by downloading the diffs from GitHub and doing git apply.

I tested the Pump Simulator (with this PR applied) on a 3rd gen SE running iOS 26.
* I was able to select Dana from the options for pumps
* I was able to modify basal rate by 0.01 U/hr increments up to the maximum allowed value of 16.0 U/hr
* I was able to bolus using 0.1 U increments
   * Bolus requests of 0.01 and 0.09 U were rejected
   * Bolus requests of 0.1, 29.9 and 30.0 were accepted
   * Loop has an overall max allowed max bolus of 30 U - I was not able to exceed 30 U for max bolus limit, so could not test Dana in-app simulator with higher bolus amounts
   * I do not think we will be exceeding 30 U without careful discussion

marionbarker (approved):
Approve from code review and test.
LGTM

PR #540, opened 2024-08-18
fix: Prevent app crash while moving presets

bastiaanv:
## Screen recording of crash:

https://github.com/user-attachments/assets/500ebf9a-6bbc-4884-9a47-c9faf9b66a4b
Screenshot 2024-08-18 at 20 24 10

Reported by @bedtime4bonzos (https://github.com/LoopKit/Loop/issues/2210)

## Proposed fix
It feels like this wasn't an issue until the override history was added. But the fix is to check what the greater allowed index is, and everything above it to cap at the original index

### DIFF
--- a/LoopKitUI/View Controllers/OverrideSelectionViewController.swift
+++ b/LoopKitUI/View Controllers/OverrideSelectionViewController.swift
@@ -388,7 +388,8 @@ public final class OverrideSelectionViewController: UICollectionViewController, 
             return originalIndexPath
         }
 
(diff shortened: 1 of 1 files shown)


bastiaanv:
Ping @ps2 & @marionbarker

marionbarker:
## Status 
Success

## Test
* Without the patch, dragging an override below the History row crashes the app. 
* After applying the patch, dragging an override below the History row positions it as the last override (above the Custom and History rows).

PR #543, opened 2024-09-10
Swedish translations

jakalm:
I noticed that there were a lot of missing Swedish translations in general for the Loop app. This PR addresses this in the LoopKit repository.

- Added missing Swedish translations in the respective `Localizable.strings` files.
- Added/Changed to `LocalizedString` calls where needed, which will help with other translations as well.
- Removed outdated translations that are no longer needed.


marionbarker:
Closing because localization is now with String Catalogs and the [lokalise](https://app.lokalise.com/projects) tool is in use for bringing in translations.

PR #544, opened 2024-09-10
Fix locale for decimal separator in Carb Entry View

jakalm:
In the latest version of Loop there is an issue introduced where it's not possible to enter decimal separator symbol for languages with a different decimal separator than a period sybmol (`.`). For example, in Swedish a comma symbol (`,`) is used.

There is also now a limit to 1 decimal number. This causes a problem when you weight something, calculate the amount of carbons in your calculator, and then end up with three decimal numbers. You are then unable to copy it in. I've increased the allowed amount of decimal numbers to three with this PR.


ps2:
I believe https://github.com/LoopKit/LoopKit/pull/546 also works, and is a little cleaner.

PR #545, opened 2024-09-21
Insulin Correction Safety and Fixes

motinis:
Please see discussion in [zulipchat](https://loop.zulipchat.com/#narrow/stream/144111-general/topic/Target.20Range/near/471619055).

This PR covers two important points:

1. The effectedSensitivity calculation needs to sum over the segments - otherwise a change in ISF can result in discarding previous segment results and thus the minimum correction calculation may be too high.
2. Calculate and apply a maxNonSuspendUnits - this is important for safety, for example in edge cases where there is a temporary higher target in play, which can result in some predicted values then having negative correction units (i.e., the weighted target is above the prediction). As a result, the final calculated dosage may result in glucose going below the suspend threshold in such regions. While not observed, it is definitely safer to implement a simple safeguard for this.

### DIFF
--- a/LoopKit/LoopAlgorithm/DoseMath.swift
+++ b/LoopKit/LoopAlgorithm/DoseMath.swift
@@ -265,12 +265,15 @@ extension Collection where Element: GlucoseValue {
         var correctingGlucose: GlucoseValue?
         var minCorrectionUnits: Double?
         var effectedSensitivityAtMinGlucose: Double?
+        
 
         // Only consider predictions within the model's effect duration
         let validDateRange = DateInterval(start: date, duration: model.effectDuration)
 
         let unit = correctionRange.unit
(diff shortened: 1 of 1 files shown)


ps2:
I think this might be covered in the LoopAlgorithm package, which will be landing soon in dev.

motinis:
@ps2 looking through the LoopAlgorithm source code I think these changes are still relevant - would you prefer I open a PR against that repo instead? Or you want to merge that to dev first and then look at this PR?

ps2:
The main part of this ticket (effectedSensitivity calculation) was fixed a while ago. The second part is maybe not possible to hit. If you can construct a realistic situation where that change is needed, please update this PR with the test (or create a new one against LoopAlgorithm, more likely)

PR #546, opened 2024-09-27
fix LoopKit/Loop#2223

Joerg-Schoemer:
accept comma in carb entry


Joerg-Schoemer:
Hi @SwiftlyNoah could you please review this little change to get the decimal separator back to work on different locales.

ps2 (changes requested):
Looks good to me generally.  Please remove the debug prints.

marionbarker:
I set my test phone location to Germany (to enable the "," on the keypad)  and Loop language to German and confirmed the problem reported in Loop Issue 2223.

I built with this modification onto the same phone.
* I confirm this fixes the problem.

PR #547, opened 2024-10-01
Proposed Fix Option 1 for Slow Build with Xcode 16

dnzxy:
Proposed solution 1 for slow build with Xcode 16 as described in https://github.com/LoopKit/Loop/issues/2235.

- Change `ENABLE_MODULE_VERIFIER` to `"$(CLANG_MODULES_ENABLE_VERIFIER_TOOL)";` which equates `No`, so disables it.


ps2:
We can disable it for debug builds.  It might be nice to save it as on for release, to at least still have some detection of issues with the module. 

Also we can disable it by setting to "No" directly. Why would you be setting it to CLANG_MODULES_ENABLE_VERIFIER_TOOL?

dnzxy:
I adopted that setting to what it was set to for existing disabled submodules; they are set to `CLANG_MODULES_ENABLE_VERIFIER_TOOL`. Frankly, not sure why it was set to that 😅but it exposes the setting to a config variable that can be set, and if not set it will default to no. So folks could easily override the setting via their ConfigOverride.xcconfig, if they wanted to ☺️ It leaves a bit more flexibility in there. Please see the linked issue for more background. 

I had left all `release` type builds set to `yes` across the various LoopWorkspace submodules but was asked to make the change for all by @marionbarker , so adopted the config variable (so `no`) for all settings.  

I‘m currently traveling. If you want this changed to a plain `no` rather than the config variable, please say so, I‘d be able to adopt that change request  next week.

marionbarker:
My rationale for requesting it be set to No for Release as well as Debug was to save 30 sec per instance with Xcode 16 (and there are 6 each verify module steps for Debug and Release).

I also did not understand why it was set for only 6 of the 19 submodules for LoopWorkspace. I figured if it was really important, it would be added to all of them.

My thinking was that the developer might want to flip it to yes, using the configuration variable, to confirm things are ok and the end users can safely have it set to no.

ps2:
My understanding of it is that it checks for some common errors in defining modules that can prevent re-use with other code. The slowdown appears to be a regression that will likely be fixed. It could be nice to have it enabled in some builds so there's a chance we see any issues with module definition, but I'm not too worried about that. I do think it would be better to use a simple "No", than adding another layer of configurability with CLANG_MODULES_ENABLE_VERIFIER_TOOL, whatever that is; I don't see that in existing code.

ps2:
> I adopted that setting to what it was set to for existing disabled submodules; they are set to CLANG_MODULES_ENABLE_VERIFIER_TOOL

@dnzxy I don't see that in the codebase when I grep for it at the root level. Where are you seeing this?

dnzxy:
I went with what Xcode suggested for me there, and used it for all instances. Would you like me to change it to just No everywhere? Can do that 👍

dnzxy:
Adjusted to No now ☺️

ps2:
Will be fixed in the `tidepool-merge` branches.

PR #548, opened 2024-10-02
Proposed Fix Option 2 for Slow Build with Xcode 16

marionbarker:
Proposed solution for slow build with Xcode 16 as described in [Loop Issue 2235](https://github.com/LoopKit/Loop/issues/2235). 

* Remove lines with `ENABLE_MODULE_VERIFIER` from `project.pbxproj` file.


marionbarker:
Closed in favor of Proposed Fix Option 1.

PR #549, opened 2024-10-05
Improve the visibility of the options to select

marionbarker:
This PR is proposed as a solution to Loop Issue 2238: [Make Checkmark List Items easier to see](https://github.com/LoopKit/Loop/issues/2238#top)

Sample screen views using for this PR.

various_dark_lineWidth_1

various_light_lineWidth_1

### DIFF
--- a/LoopKitUI/Views/CheckmarkListItem.swift
+++ b/LoopKitUI/Views/CheckmarkListItem.swift
@@ -74,7 +74,7 @@ public struct CheckmarkListItem: View {
                 .foregroundColor(.accentColor)
         } else {
             Circle()
-                .stroke()
+                .stroke(Color.accentColor, lineWidth: 1)
                 .foregroundColor(Color(.systemGray4))
         }
     }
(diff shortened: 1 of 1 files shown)


dnzxy:
May I suggest to either
- conditionally (for dark mode only) adjust the color of the control, i.e., a lighter gray, or extending the line width, or 
- opting to use white or black color, depending on theme, so `.primary`.

The blue color (`.accentColor`) is associated with something active / selected.

marionbarker:
In the Issue, @ps2 requested "stick to existing iOS conventions for color".

I was not quite sure how to do that:

* Increasing lineWidth without making color modifications was not sufficient
* Replacing Color.accent with Color.primary gave good contrast but was more black and white than @ps2 example graphics
* I adjusted the .foregroundColor from systemGray4 to systemGray2 to systemGray
    * I thought that looked better, but I could hear @itsmojo asking me to consider his "aging & colorblind eyes"
    * So I also increased lineWidth to 2

I captured a bunch of other screen shots, but I think these are sufficient.

Original and current commit Dark

original_revised_dark

Original and current commit Light

original_revised_light

itsmojo:
I'm not qualified to say anything on color schemes, but the current commit Dark and Light views do look good to my eyes...

ps2:
Yes, I think the line width increase makes it a lot more visible, and the color matches the conventions we see in iOS.  Thanks!

PR #552, opened 2024-10-19
Decrease lower bound of correction range to 75

sterrell586:


marionbarker:
It looks like these changes are a personal customizations rather than updates intended for the wider Loop community. 

This is a common occurrence, so no worries! But please close this PR.

BR/Marion

PR #556, opened 2025-02-12
Fix Decay Effect to be a continuous function

motinis:
This fixes the LoopMath  implementation to ensure that the decay effect function is continuous and thus doesn't show discontinuity effects related to user sensor time (as reported in https://github.com/LoopKit/Loop/issues/2287). The test demonstrates that shifting a glucose reading by 1 microsecond earlier results in a less than 1E-6 change to glucose prediction, while the prior implementation would have resulted in a massive jump at the boundary (i.e. after 1 microsecond elapsed).

Subsequent fixes would be necessary to be done to the LoopDataManagerDosing tests which rely on fixtures which are calibrated with the old implementation which was inaccurate when sensor readings were not extremely close to the beginning of 5 minute boundaries.

### DIFF
--- a/LoopKit/LoopMath.swift
+++ b/LoopKit/LoopMath.swift
@@ -193,20 +193,27 @@ extension GlucoseValue {
 
         let glucoseUnit = HKUnit.milligramsPerDeciliter
         let velocityUnit = GlucoseEffectVelocity.perSecondUnit
-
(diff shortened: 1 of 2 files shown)


motinis:
@ps2 - could you please take a look at this?

PR #557, opened 2025-04-11
Fix carb entry bug, Issue 2307

marionbarker:
## Purpose

Fix for [Loop Issue 2307](https://github.com/LoopKit/Loop/issues/2307).

## Method

I asked for help on solving this and got a suggestion from @bjorkert (thanks).

Once the file that needed to be modified was identified, I checked the status of that file in the `tidepool-merge` branch, `LoopKit/LoopKitUI/CarbKit/CarbQuantityRow.swift`
* The file has been updated, so I started with the `tidepool-merge` version
   * The bug still happens with that version of the file
* I modified the suggestion from @bjorkert to work with the `tidepool-merge` version
* I configured a `minAllowedCarbEntry` of 1.0 (instead of using hard-coded 1.0) and required the entry be >= `minAllowedCarbEntry` before the continue button is enabled

I'm happy to modify this as desired.

## Test

This fixes the reported bug:
* A user can no longer type  ".", "0", "5" and wind up with 5 g of carbs
* The entered value must be >= 1 before the user can continue
* If a value < 1 is entered, the continue button is inactive

### DIFF
--- a/LoopKitUI/CarbKit/CarbQuantityRow.swift
+++ b/LoopKitUI/CarbKit/CarbQuantityRow.swift
@@ -76,7 +76,9 @@ public struct CarbQuantityRow: View {
     
     // Update quantity based on text field input
     private func updateQuantity(with input: String) {
-        let filtered = input.filter { "0123456789.".contains($0) }
+        let decimalSeparator = formatter.decimalSeparator ?? "."
+        let allowedCharacters = "0123456789" + decimalSeparator
+        let filtered = input.filter { allowedCharacters.contains($0) }
         if filtered != input {
             self.carbInput = filtered
         }
--- a/LoopKitUI/CarbKit/CarbQuantityRow.swift
+++ b/LoopKitUI/CarbKit/CarbQuantityRow.swift
@@ -22,7 +22,6 @@ public struct CarbQuantityRow: View {
     private let formatter: NumberFormatter = {
         let formatter = NumberFormatter()
         formatter.numberStyle = .decimal
-        formatter.maximumIntegerDigits = 3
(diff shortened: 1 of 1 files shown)


bjorkert:
This version does not allow "," as decimal separator. If I enter "0,5", I get "5".

marionbarker:
I tested this with 2 phones. One set for "." as a decimal separator and one for "," as a decimal separator.

It does prevent the case of ",05" or ".05" being interpreted as "5" g for both types of separators.

There is a difference in behavior depending on the separator:
* for "," as decimalSeparator: the continue button is disabled for any entry that includes a ","
   * In other words, 10 will be accepted, "10," will not
* for "." as decimalSeparator: the continue button is enabled for any entry >= 1 whether decimal place is included
   * In other words, 10.5 is accepted and displays as 10, 10.9 becomes 11

LoopKitUI/CarbKit/CarbQuantityRow.swift
```diff
@@ -77,19 +76,26 @@ public struct CarbQuantityRow: View {
     
     // Update quantity based on text field input
     private func updateQuantity(with input: String) {
-        if let number = formatter.number(from: input) {
-            quantity = number.doubleValue
+        let decimalSeparator = formatter.decimalSeparator ?? "."
+        let allowedCharacters = "0123456789" + decimalSeparator
```
ps2, line 80: Has anyone tested that this works for arabic/persian locales?

ps2 (commented):
Not great from a locale-supporting standpoint, but I see the separatory is covered.

PR #560, opened 2025-05-07
Update LoopKit tidepool-merge with improvements from dev

marionbarker:
## Purpose

Bring in improvements added to the dev branch since the tidepool-merge branch was first pushed to LoopKit

## Test

Start with a local clone using LoopWorkspace `tidepool-merge` branch:

* Perform a test merge of main into tidepool-merge for LoopKit in that local clone
* Open Xcode workspace and build with modified LoopKit

Confirm the checkmark limit is easy to see and Carb input accepts comma as a decimal separator so long as the leading zero is typed.
The issue pointed out in Loop 2307 is still present but will be corrected in a separate PR.


ps2 (approved):
LGTM!

PR #573, opened 2025-10-20
ios26 fixes

ps2:
* DIsabled actions buttons look disabled
* Information views do not auto-dismiss
* Saving values in warning ranges works again


marionbarker (approved):
Approve based on test. See https://github.com/LoopKit/Loop/pull/2371#issuecomment-3424184868 which indicates failures described in #2360 are fixed.

~~I don't have an example where the inactive button was not grey and where information views auto-dismissed, so I can't demonstrate these are fixed. But willing to approve and accept this change without that test.~~

I did test examples for all 3 fixes as reported in later comments under Loop PR 2371.

PR #574, opened 2025-10-22
Protect against invalid immutable boluses with future end date

ps2:


marionbarker:
## Test

I'm not sure these tests exercise what this modification is supposed to address.
I believe updates are planned to DanaKit and MedtrumKit.

### Reproduce DanaKit error

First build with release/3.8.2, SHA `e02e46a` and attach a Dana pump using an rPi simulator
Issue a large bolus and then ^C out of the rPi simulator
Observe reported IOB for Loop main screen

1. Attach Dana
2. wait for green loop
3. bolus 5 U: before IOB = 0.06, after 5.06
4. control C out of simulator - the bolused amount halted at 0.91 U of 5.00 U - stuck on that reading with tap to stop showing (do not touch it)
5. wait a few loop cycles
6. Showing a yellow loop (no connection to pump), but IOB is reduced to 1.07 U

### DanaKit error still present

> Note - I think this is the wrong test for exercising code changes in the PR. 

Modify the LoopKit submodule to ps2/invalid-bolus-protection and build onto the same test phone.

1. After the new build is installed, restart the Dana simulator (use same status.json)
2. Get a green loop, IOB is still too low (now 0.95 U)
3. bolus 3 U: before IOB = 0.95, after 3.94
4. control C out of simulator - the bolused amount halted at 1.14 U of 3.00 U - stuck on that reading with tap to stop showing (do not touch it)
5. wait a few loop cycles
6. IOB reduced to 2.08

There is no change here. I don't know how much might be associated with this being a simulator.

I also know that @bastiaanv is planning updates to DanaKit (and MedtrumKit) code.

### DanaKit error still present

> Note - I think this is the wrong test for exercising code changes in the PR. 

This time, test by having the pump continue operating on the rPi, but take the phone out of range of the rPi after the bolus.

1. restart the rPi simulator, wait for green loop
2. IOB reports 2.08 U
3. bolus 3 U: before IOB = 2.08, after 5.07
4. walk the phone out of range
5. loses signal and the bolus progress display stopped displaying (I think around 2 U delivered)
6. wait a few minutes and recover the phone
7. shows pump signal loss and IOB of 4.13
   * check bolus history
   * interrupted bolus of 2.11 of 3 at 11:01 (phone out of range
   * interrupted bolus of 1.14 of 3 at 10:51 (this is ^C of rPi)
   * interrupted bolus of 0.91 of 5 at 10:38 (this is ^C of rPi)

marionbarker (approved):
Tested that this builds and runs.

PR #577, opened 2025-10-31
Revert pumpevent fix attempt

ps2:
Modifying the events pumpmanagers report was hack to try to deal with invalid events. Fixing the pumpmanagers is a better solution.  The pumpmanagers that were reporting the invalid events are removed from the main code now.

Adding a fatalError if invalid pump events are submitted, so pumpmanager developers will discover issues like this faster.

### DIFF
--- a/LoopKit/InsulinKit/DoseStore.swift
+++ b/LoopKit/InsulinKit/DoseStore.swift
@@ -684,6 +684,12 @@ extension DoseStore {
             return
         }
 
+        for event in events {
+            if let dose = event.dose {
+                self.log.debug("Add %@, isMutable=%@", String(describing: dose), String(describing: event.dose?.isMutable))
+            }
+        }
+
         persistenceController.managedObjectContext.perform {
             var lastFinalDate: Date?
             var firstMutableDate: Date?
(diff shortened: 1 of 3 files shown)


marionbarker:
## Test

### Test Steps Definition

The most reliable test for this issue is:
1. Pair a new DASH pod
2. Attempt to add a carb entry
   * If that fails, the error is present
   * If that succeeds, the error is not present

### Configuration

Begin with release/3.8.2, commit `c885710`, "update LoopKit: restore commit `2a1dfc2e`;"
* Perform Test Steps - no error

### Test 1

Modify LoopKit to `ps2/revert-pumpevent-fix-attempt` branch, commit `3a677573`, 
* Perform Test Steps - no error

### Test 2

Step back 1 commit in LoopKit (to `6921a00e`) - which provides a fatal error if a pump manager has a particular bug in it.
* Perform Test Steps - crash happens and is revealed to be triggered from OmniBLE as part of the Insert Cannula steps
* Do a quit/restart (rebuild the app) and can now insert the cannula
* Carbs can be saved without the error

### Future steps

Once pump managers do not trigger this error, then the fatal error, if it happens, can be added back to assist in identifying this issue for any future new pump managers.

marionbarker (approved):
Approved from code review and test.
LGTM

Issue #531, opened 2024-04-15
Please give Lyumjev better DIA/PT/ED values before launching Loop Main 3.4.0?


oddst:
I really believe that Loop developers should change the current Lyumjev implementation from the very unfortunate idea that Lyumjev=FIASP, to giving Loop the values initially recommended by Dragan and confirmed by many members of LnL and most loopers in Norway using Lyumjev. I am aware of the previous argumentation that these values are not really that important, but seriously that argument would be the same as arguing that Loop in itself is pointless. Making these decisions less correct 288 times every day means that even smaller errors gets big when summarized. Yes, people have different metabolism, but that still does not mean that we should use numbers that is different from manufacturer data and what has been preferred by lots of experienced users.

Changing this now should not be argued against due to risk, since that would be the same as admitting that Loop is using bad values for Lyumjev. It seems very likely that this change will improve the situation for users (if using default values from Loop).

There are many users of Loop that have been using these recommended settings for Lyumjev, and with really good results.

It is about time that Loop changes the previous verdict to treat Lyumjev second-rate by not implementing more correct numbers.

Please do this before launching Loop Main 3.4.0, as this DIY environment otherwise will force a huge number of loopers to have to implement “individual patching” using webbrowser building with the much increased complexity that gives…!

DIA=300 (from 360)
PT=62    (from 55)
ED=5      (from 10)

motinis:
@oddst just wondering if you could provide some more information on that validation? Should it be equally applicable to children too?

oddst:
In my opinion, Loop developers did a clear mistake when they chose to treat Lyumjev = FIASP. When Lyumjev support was introduced in Loop, this shortcut was done by mistake. If you choose to say that Lyumjev works like FIASP, then you could similarly say that there are no need to treat Humalog og Novorapid/Novolog any different either.

Loop was created with an idea about more precise calculations than with PDM/Controllers very course calculations. The idea was that erronous settings and insulin absorption curves effect the solution much more negatively due to calculations every 5 minutes. 

But with the introduction of Lyumjev, unfortunately Loop developers argued the opposite - that the 5 minute calculations made in unnecessary to have values that better suited Lyumjev. 

I originally asked the creator of today's insulin and carbs models in Loop (Dragan) if he could look at how Lyumjev best could be introduced in Loop. His response was the suggestions above. 

These suggestions was also the consensus from other early users of Lyumjev as well. 

[...]

PR #580, opened 2025-12-03
Update lyumjev peak to 45 minutes

marionbarker:
## Purpose

Modify the lyumjev peak from 55 minutes to 45 minutes.

The lyumjev exponential model parameters match those of the fiasp model (code before this PR)
* See my comment in [LoopKit Issue 531](https://github.com/LoopKit/LoopKit/issues/531#issuecomment-3348633407)
* This issue was primarily focused for Loop Users
* Now, these insulin models are also used for Trio which modifies the exponential model:
   * In Trio, the actionDuration for all insulins is modified to a default value of 10 hours
   * The longer actionDuration modifies the shape of the curve in the early hours following delivery
* The request from Trio developers is to change the peakActivity from 55 to 45 minutes
   * This simplifies the use of lyumjev for Trio users in that they no longer need to manually modify that parameter

## Consequence

If this is accepted, then it will be announced that anyone who currently uses the lyumjev model needs to be aware that the peak time is different. If they got better results with Loop before the change, they can select the fiasp model instead.

The [comment](https://github.com/LoopKit/LoopKit/issues/531#issuecomment-3348890351) from @damonbayer is that Loop users will not notice much difference.

Evidently Trio users are affected more by this change.

### DIFF
--- a/LoopKit/Insulin/ExponentialInsulinModelPreset.swift
+++ b/LoopKit/Insulin/ExponentialInsulinModelPreset.swift
@@ -42,7 +42,7 @@ extension ExponentialInsulinModelPreset {
         case .fiasp:
             return .minutes(55)
         case .lyumjev:
-            return .minutes(55)
+            return .minutes(45)
         case.afrezza:
             return .minutes(29)
         }


dnzxy:
Just wanted to add here: while EMA's official publications for PK and PD seem to point to ~57min (so probably closer in line with 55min), a 45min peak for Lyumjev has been the default for AndroidAPS' exponential insulin modeling for quite a while, and we have some publications/data that also show earlier than 55min peak. It's sensible change, so we definitely welcome it – irrespective of what we model wrt. DIA or a user-customizable peak in Trio beyond the default peak time dictated by insulin selection.

ps2:
@dnzxy Could you add links for "publications/data that also show earlier than 55min peak"?

dnzxy:
- Profil (presented at ATTD 2020 in Madrid) discussed this; no link to that, sorry.
- This points to an earlier "bulk" of insulin action, interpreted as effective peak being earlier than the EMA labelled 57 min. 
https://www.ovid.com/journals/domet/fulltext/10.1111/dom.14094~ultra-rapid-lispro-lowers-postprandial-glucose-and-more
- There are several publications out of the OSAID community around this topic. David Burren, Tim Street, and Bernd Herppichböhm have written on the matter
  - https://www.diabettech.com/lyumjev/insulin-lyumjev-not-as-urli-as-i-expected/
  - https://bionicwookiee.com/2021/12/26/lyumjev/
  - https://www.diabettech.com/wp-content/uploads/2020/10/Lyumjev_w.UAM-week_Be_13Oct20.pdf
- As pointed out earlier, AndroidAPS uses a 45min peak. Trio users also are advised and typically use a custom peak of 45min when using Lyumjev. I suspect that’s be n in the multiple hundreds, if not thousands :) 

Edited to add a few links and also to point to this discussion on Loop Zulip where also lower peak times where discussed.

ps2:
Thanks for adding links. Reading through those, it seems like all the observed curves show later times, like 55 minutes or more for peak? 

I'm not surprised that people are having ok results with an earlier peak. These details on timing are pretty small when all the other sources of noise/variation are considered, and the effect on dosing is very small. But that's not much of an argument for changing it; the argument that people aren't having problems with an earlier peak. 

Seems like the best evidence for what we should actually expect for effects is showing that 55 or later is probably more appropriate.  Or am I missing something?

dnzxy:
> Thanks for adding links. Reading through those, it seems like all the observed curves show later times, like 55 minutes or more for peak?

Anytime :) I'm not really sure how you are reading later times out of them. All I can say, is that a majority of the OS-AID community (the oref portion) has been running with a Lyumjev peak time of 45min for a long time, so it would be great to see that adjusted/updated.

From reading over the initial insulin modeling issue and PR it seems like (I am deliberately paraphrasing very casually, no offense to Damon or anyone else) it was a "well, for Lyum, no idea man, let's just go with Fiasp, it'll be okay" and that's what's been implemented over the years. In a more recent comment, linked by Marion above, he also said that while there may not be a lot of proof from what he can tell for a lowered peak, there also is not really much for the current status quo — so from where I am standing, it makes sense to take a step back, look at this from a perspective of "the entire OS-AID ecosystem" and align the peak times across the operating systems.

Open to discussing pro's and con's, as always. 

PS: thanks for enganging in this Pete :)

damonbayer:
> "well, for Lyum, no idea man, let's just go with Fiasp, it'll be okay" and that's what's been implemented over the years. In a more recent comment, linked by Marion above, he also said that while there may not be a lot of proof from what he can tell for a lowered peak, there also is not really much for the current status quo — so from where I am standing, it makes sense to take a step back, look at this from a perspective of "the entire OS-AID ecosystem" and align the peak times across the operating systems.

This is a fair characterization of my view. For what it's worth, I have used both Lyumjev and Fiasp with Loop and didn't feel like the insulin absorption model caused any issues.

> * In Trio, the actionDuration for all insulins is modified to a default value of 10 hours

If Trio is able to modify the default `actionDuration` parameter, I don't see why they can't also modify the default peak parameter. Why do users have to change it themselves?

dnzxy:
@damonbayer because there‘s a difference in how oref systems treat insulin peak and DIA. DIA is part of the therapy profile, and we just set the default value to something different; it can still be customized by the user. We don’t default the custom peak setting to true and immediately set it based on insulin selection. That’d be a different thing, but is probably also not really important here. 

Trio is not really the point here, tbf, having a peak that aligns with what is used across systems is. 

You can make the argument that this would be consistency out of principle, that is fair. It’s still worth doing in our view, which is why we asked Marion to PR it :)

tim2000s:
Hi guys, I think your holding the peak time to 55 minutes plus based on data that links to clamp studies misses the point about dosing sizes in AID systems, and the resultant effective peak size. 

I've run an observational study against multiple users on Trio and AAPS and the smaller average SMB size points at a peak much closer to 40-45 minutes than the large sizes used in the clamps. 

Paper attached. 
[insulin-kinetics-paper-2026-08-05.pdf](https://github.com/user-attachments/files/30761233/insulin-kinetics-paper-2026-08-05.pdf)

damonbayer:
@tim2000s The typical observed peak being less than 55 minutes is notable, but we also see estimates as low as 8 minutes and as high as 112 minutes. Makes me skeptical of the quality of the methodology, as does the very llm-sounding prose.

PR #586, opened 2026-02-11
Pretty print pump events

marionbarker:
## Purpose

This is in response to [Loop Issue 2383](https://github.com/LoopKit/Loop/issues/2383).

This PR along with [Loop PR 2403](https://github.com/LoopKit/Loop/pull/2403) modify the formatting for the pump event history so the detailed view is easier to read.

### DIFF
--- a/LoopKit/InsulinKit/DoseEntry.swift
+++ b/LoopKit/InsulinKit/DoseEntry.swift
@@ -155,16 +155,16 @@ extension DoseEntry {
     public var formatted: String {
         return [
             "type: \(type)",
-            "isMutable: \(isMutable)",
             "automatic: \(optionalString(automatic))",
-            "startDate: \(startDate)",
-            "endDate: \(endDate)",
+            "isMutable: \(isMutable)",
             "deliveredUnits: \(deliveredUnits != nil ? deliveredUnits!.insulinFormatter : "nil")",
             "units: \(value.insulinFormatter)",
-            "description: \(optionalString(description))",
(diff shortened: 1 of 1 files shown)


marionbarker:
Successful test reported at this link: https://github.com/LoopKit/Loop/pull/2403#issuecomment-3885738546

LoopKit/InsulinKit/DoseEntry.swift
```diff
@@ -276,3 +294,25 @@ extension DoseEntry: RawRepresentable {
...
+    var insulinFormatter: String {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .decimal
+        formatter.minimumFractionDigits = 1
+        formatter.maximumFractionDigits = 3
+        formatter.locale = Locale.current
+        if let formattedValue = formatter.string(from: self) {
+            return(formattedValue)
+        } else {
+            return("nil")
```
itsmojo: How about rewriting these two returns as`return formattedValue` and `return "nil"` (having a space after the keyword `return` and not using parentheses) to match other places in this file and in the vast majority of Loop.

itsmojo (approved):
It would be good to fix the formatting nit, but obviously not a showstopper

ps2 (approved):
LGTM!

LoopKit/InsulinKit/DoseEntry.swift
marionbarker: Done.

PR #588, opened 2026-02-22
Calculate Insulin Effects accurately by taking into account ISF according to the schedule

motinis:
Loop has historically used the ISF at the start of a dose to determine the ISF for the lifetime of the dose, ignoring changes to ISF (whether according to the regular schedule or via overrides).

This contribution changes this to using ISF over the lifetime of the dose. It does this by splitting the calculation in 2 parts:

1. The drop in IOB for a given dose over time
2. Multiplying the IOB drop for each "bin" by the constant ISF for that bin (where bins are typically 5 minutes, but can be less, if there happens to be a change in ISF).

Dosage determination is handled by considering the glucose effect change that would happen if giving 1U, and then scaling this to match Loop's existing criteria.

Note that the interpolation during dose determination is done for consistency, even though unitGlucoseEffect should be aligned on the same dates. The interpolation logic was originally implemented in https://github.com/LoopKit/LoopKit/pull/562 before being generalized in https://github.com/motinis/LoopKit/commit/3040b7be0b46926a6429e2473dc2205ccfeec922  and simplified for 1 value - it is designed to support interpolation as part of a loop where we want to continue to search for the next relevant data point(s) from where we left off at the previous iteration (so as to have O(n) behavior in the length of the data and not O(n^2)).


LoopKit/LoopAlgorithm/DoseMath.swift
```diff
@@ -230,54 +230,32 @@ extension Collection where Element: GlucoseValue {
...
         let validDateRange = DateInterval(start: date, duration: model.effectDuration)
 
         let unit = correctionRange.unit
         let suspendThresholdValue = suspendThreshold.doubleValue(for: unit)
+        
+        let unitEffects = [DoseEntry(type: .bolus, startDate: date, value: 1, unit: .units)].glucoseEffects(insulinModelProvider: StaticInsulinModelProvider(model), longestEffectDuration: model.effectDuration, insulinSensitivity: sensitivity)
+
+        var unitEffectsIndex = 0
 
         // For each prediction above target, determine the amount of insulin necessary to correct glucose based on the modeled effectiveness of the insulin at that time
```
oliverstory, line 252: This comment should now read something like:
// For each prediction above target, determine the amount of insulin necessary to correct glucose based on the modeled effectiveness of insulin over the dose duration

LoopKit/LoopAlgorithm/DoseMath.swift
oliverstory, line 226: This comment needs updating as 'sensitivity' is now the insulin sensitivity schedule (not sensitivity at the time of delivery)

oliverstory (commented):
See comments on lines 226 and 252 of DoseMath.swift, where code comments need updating due to the switch from using ISF at the time of a dose to using the ISF schedule.

ps2:
Mid-absorption ISF is part of the changes landing with the tidepool merge, and LoopAlgorithm package.

PR #605, opened 2026-08-20
Xcode 27 Fix

Camji55:


LoopKitUI/Views/Presets/CreatePresetNameAndScheduledEdit.swift
```diff
@@ -38,7 +38,7 @@ struct CreatePresetNameAndScheduledEdit: View {
     @Binding var preset: NewCustomPreset
     @Binding var path: NavigationPath
     
-    @State private var isDurationPickerExpanded = false
+    @State private var isDurationPickerExpanded: Bool
```
Camji55, line 41: `false` is the default for this property in the `init` below. Declaring it here is redundant and produces a build error in Xcode 27

